Issue #12, opened 2023-06-28
SPI mode and bit order not set correctly


trevor-makes:
See [this line in SPI.cpp](https://github.com/arduino/ArduinoCore-renesas/blob/149f78b6490ccbafeb420f68919c381a5bdb6e21/libraries/SPI/SPI.cpp#L386C38-L386C38)

`spcmd0 &= !(((uint32_t)0xFF) << 2);` should be `spcmd0 &= ~(((uint32_t)0xFF) << 2);`

The logical not makes the mask 0 and wipes out the earlier settings. Should be a bitwise not to make the mask FFFFFC03.

trevor-makes:
Looking into this a bit further, I'm not sure why this would be the mask setting. BRDV is only two bits (b3, b2), so I think the mask should actually be `~(((uint32_t)0x03) << 2)`.

I'm experimenting with setting the registers directly, but haven't had any luck thus far.

trevor-makes:
Aha, got it working... I was using the SPI0 registers, but should have been using SPI1. Adding the following after calling `beginTransaction` set the expected LSB and MODE3 behavior, so fixing the mask as above should resolve the issue.

```
  R_SPI1->SPCMD_b[0].CPHA = 1;
  R_SPI1->SPCMD_b[0].CPOL = 1;
  R_SPI1->SPCMD_b[0].LSBF = 1;
```

aentinger:
Closing this issue since it is heavily outdated, see https://github.com/arduino/ArduinoCore-renesas/pull/45 .

PR #13, opened 2023-06-28
Fix SPI mode and bit order settings

trevor-makes:
#12 
The logical not in [this line](https://github.com/arduino/ArduinoCore-renesas/blob/149f78b6490ccbafeb420f68919c381a5bdb6e21/libraries/SPI/SPI.cpp#L386C38-L386C38) makes the mask 0 and wipes out the earlier settings. This should be a bitwise not, but even so the `0xFF` would also mask off the SSLA bits b6-b4 and half of the SPB bits b9-b8. Changing `0xFF` to `3` will clear just the BRDV bits b3-b2.


MrYsLab:
I verified the fix using an ADXL345, and it works!
Thanks.

facchinm:
Thank you so much for the fix!

PR #16, opened 2023-06-30
Fix RX_LED #define for R4 Minima

iadjedj:
RX_LED is connected to Port0-13 on the Minima

image

R4 WiFi probably also needs a fix (the defines for RX_LED/TX_LED are the same)
But I did not dig on how the firmware handles RX/TX with the ESP32

See : https://github.com/arduino/ArduinoCore-renesas/blob/main/variants/UNOWIFIR4/includes/ra_cfg/fsp_cfg/bsp/bsp_pin_cfg.h


facchinm (changes requested):
The right place to add this modification is https://github.com/arduino/ArduinoCore-renesas/blob/main/variants/MINIMA/pins_arduino.h#L73-L76 since this file is autogenerated and can't be used directly.

facchinm:
Merged as https://github.com/arduino/ArduinoCore-renesas/pull/67

PR #19, opened 2023-06-30
Fix undefined behavior

alrvid:
Fixes undefined behavior where SDCardBlockDevice::wait_for_completition() runs off the end without returning a value.


alrvid:
Also fixes an overflow in the reported size of block devices by changing the data type to 64 bits and by preventing a multiplication overflow.

PR #30, opened 2023-07-04
Update RTC_Alarm.ino

karlsoderby:
Update RTC_Alarm to work with UNO R4 Minima & WiFi boards.

Changes made.
- Switch the LED used (from RGB > LED_BUILTIN). R4 boards have no RGB so this example does not compile/work.
- Made `ledState` global so that it can be used in `loop()` as well.
- Added serial functions to the loop to print the current state of the alarm. Alarm state is now printed each second.
- Moved `alarmCallback()` function to the bottom for better readability.

Example is tested and verified on both UNO R4 boards and the Portenta C33.


aentinger (approved):
I can confirm it compiles for Uno R4 WiFi, Minima and Portenta C33. LGTM :+1:

PR #32, opened 2023-07-04
Add LED Matrix Default Example

karlsoderby:
This adds the `MatrixIntro` example to the list of examples for the Arduino_LED_Matrix library.


aentinger (approved):
Compiles, LGTM :+1:

PR #42, opened 2023-07-10
LED_matrix: move _sequenceDone = true out of if block

KurtE:
Currently if you write a sketch that relies on knowing when a sequence completes, by having your code call sequenceDone(), this will only work if your sketch sets a non-null callback function, as the code that sets the condition is currently in the if block:
```
if(_callBack != nullptr){
```
Simply moved the line to after the if block

### DIFF
--- a/libraries/Arduino_LED_Matrix/src/Arduino_LED_Matrix.h
+++ b/libraries/Arduino_LED_Matrix/src/Arduino_LED_Matrix.h
@@ -175,8 +175,8 @@ public:
             }
             if(_callBack != nullptr){
                 _callBack();
-                _sequenceDone = true;
             }
+            _sequenceDone = true;
         }
         memcpy(framebuffer, (uint32_t*)frame, sizeof(frame));
     }


KurtE:
> [CLA assistant check](https://cla-assistant.io/arduino/ArduinoCore-renesas?pullRequest=42) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/arduino/ArduinoCore-renesas?pullRequest=42) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/arduino/ArduinoCore-renesas?pullRequest=42) it.

Sorry, I am trying to decide if I will sign such agreement, that one should probably have an attorney review it.   Things like:

"By signing in via GitHub acknowledge and agree that (A) your Contribution is your original work."

What does "Original" mean?  I simply fixed a problem in the current code base. Is that Original?

But I will think about it.

KurtE:
Signed, but may want to put a hold on this.  Looks like the system picked up the updates to all of the links that had to be patched

Will try to update it to just have the one file with the simple change

KurtE:
Fixed to only update the one file

facchinm:
Merged, thanks you very much!

PR #51, opened 2023-07-14
Use library for NTP data handling

sebromero:
This is a simplified version of the RTC NTP sync example that uses an existing library to handle everything NTP instead of reinventing the wheel. I'm not sure if we want to avoid having such dependencies. Any feedback on this @facchinm ?


aentinger (approved):
Not reinventing the wheel definitely makes sense. LGTM :+1: @sebromero .

PR #52, opened 2023-07-16
Bugfix: `micros()` returns smaller number when timer wraps

Kees-van-der-Oord:
See [issue 49](https://github.com/arduino/ArduinoCore-renesas/issues/49)
1. Fix for anomaly in micros() when the counter wrapped just before it was read (same solution as in R3 code)
2. Specify timer configuration as constants to make micros() considerably faster

### DIFF
--- a/cores/arduino/time.cpp
+++ b/cores/arduino/time.cpp
@@ -2,8 +2,10 @@
 #include "IRQManager.h"
 #include "FspTimer.h"
 
+// this file implements the following public funcions: delay, delayMicroseconds, yield, millis, micros
+// The millis and micros implementation uses timer AGT0 (24 HMz, 16-bits, count-down mode, 1 ms period)
+
 volatile unsigned long agt_time_ms = 0;
-uint32_t _freq_hz = 0;
 
 __attribute__((weak)) void delay(uint32_t ms) {
 	R_BSP_SoftwareDelay(ms, BSP_DELAY_UNITS_MILLISECONDS);
@@ -17,20 +19,24 @@ __attribute__((weak)) void yield() {
 }
 
 static FspTimer main_timer;
-
-static uint32_t _top_counter;
+// specifying these details as constants makes micros() faster !
+#define _timer_type           AGT_TIMER
+#define _timer_index          0
+#define _timer_underflow_bit  R_AGT0->AGTCR_b.TUNDF
+#define _timer_clock_divider  TIMER_SOURCE_DIV_8 // dividers 1, 2, 4 and 8 work because _timer_period is < 16-bit. bug in R4 1.0.2: divider 4 acts as 1 !? 
+#define _timer_clock_freq     24000000UL
+#define _timer_ticks_per_us   (_timer_clock_freq / ((1 << _timer_clock_divider) * 1000000UL))
+#define _timer_period         (_timer_ticks_per_us * 1000UL)
 
 static void timer_micros_callback(timer_callback_args_t __attribute((unused)) *p_args) {
-	agt_time_ms += 1; //1ms
+	agt_time_ms += 1;
 }
 
 void startAgt() {
-	main_timer.begin(TIMER_MODE_PERIODIC, AGT_TIMER, 0, 2000.0f, 0.5f, timer_micros_callback);
-	IRQManager::getInstance().addPeripheral(IRQ_AGT,(void*)main_timer.get_cfg());
+	main_timer.begin(TIMER_MODE_PERIODIC, _timer_type, _timer_index, _timer_period, 1, _timer_clock_divider, timer_micros_callback);;
+	main_timer.setup_overflow_irq();
 	main_timer.open();
-	_top_counter = main_timer.get_counter();
-	main_timer.start();
-	_freq_hz = main_timer.get_freq_hz();
+	main_timer.start(); // bug in R4 1.0.2: calling start() is not necessary: open() starts the counter already !?
 }
 
 unsigned long millis()
@@ -43,10 +49,11 @@ unsigned long millis()
 }
 
 unsigned long micros() {
-
-	// Convert time to us
+	// Return time in us
 	NVIC_DisableIRQ(main_timer.get_cfg()->cycle_end_irq);
-	uint32_t time_us = ((main_timer.get_period_raw() - main_timer.get_counter()) * 1000 / main_timer.get_period_raw()) +  (agt_time_ms * 1000);
+	uint32_t ms = agt_time_ms;
+	uint32_t cnt = main_timer.get_counter();
+	if (_timer_underflow_bit && (cnt > (_timer_period / 2))) ++ms; // the counter wrapped arround just before it was read
 	NVIC_EnableIRQ(main_timer.get_cfg()->cycle_end_irq);
-	return time_us;
+	return  ms * 1000 + (((_timer_period - 1) - cnt) / _timer_ticks_per_us);
 }
--- a/cores/arduino/time.cpp
+++ b/cores/arduino/time.cpp
@@ -28,8 +28,8 @@ void startAgt() {
 	// in 1 ms (period) is an integer number and below the 16-bit counter limit
 	// on the Uno R4 the AGT clock is 24 MHz / 8 -> 3000 ticks per ms
 	// on the Portenta C33 the AGT clock is 50 Mhz / 8 -> 6250 ticks per ms
-	const uint32_t clock_freq = R_FSP_SystemClockHzGet(FSP_PRIV_CLOCK_PCLKB);
-	const uint32_t period = clock_freq / ((1 << TIMER_SOURCE_DIV_8) * 1000UL);
+	const uint32_t clock_freq_Hz = R_FSP_SystemClockHzGet(FSP_PRIV_CLOCK_PCLKB);
+	const uint32_t period = clock_freq_Hz / ((1 << TIMER_SOURCE_DIV_8) * 1000UL);
 	agt_timer.begin(TIMER_MODE_PERIODIC, AGT_TIMER, 0, period, 1, TIMER_SOURCE_DIV_8, timer_micros_callback);;
 	agt_timer.setup_overflow_irq(8);
 	agt_timer.open();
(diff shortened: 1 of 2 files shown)


aentinger:
This fixes #49.

aentinger (changes requested):
Hi @Kees-van-der-Oord :coffee: :wave: 

Thank you for your PR. I'm inclined to merge it but there are a couple of suggestions I'd like you to tackle first.

cores/arduino/time.cpp
```diff
@@ -43,10 +50,11 @@ unsigned long millis()
 }
 
 unsigned long micros() {
-
-	// Convert time to us
+	// Return time in us
 	NVIC_DisableIRQ(main_timer.get_cfg()->cycle_end_irq);
-	uint32_t time_us = ((main_timer.get_period_raw() - main_timer.get_counter()) * 1000 / main_timer.get_period_raw()) +  (agt_time_ms * 1000);
+	uint32_t ms = agt_time_ms;
```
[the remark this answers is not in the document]
Kees-van-der-Oord: The variable ms is incremented when the timer wrapped, so it cannot be const

cores/arduino/time.cpp
```diff
@@ -43,10 +50,11 @@ unsigned long millis()
...
+	uint32_t cnt = main_timer.get_counter();
+	if (_timer_underflow_bit && (cnt > (_timer_period / 2))) ++ms; // the counter wrapped arround just before it was read
 	NVIC_EnableIRQ(main_timer.get_cfg()->cycle_end_irq);
-	return time_us;
+	return  ms * 1000 + (((_timer_period - 1) - cnt) / _timer_ticks_per_us);
```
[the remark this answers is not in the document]
Kees-van-der-Oord: I will break it down for more clarity

cores/arduino/time.cpp
```diff
@@ -17,20 +19,25 @@ __attribute__((weak)) void yield() {
 }
 
 static FspTimer main_timer;
+// specifying these details as constants makes micros() faster !
+#define _timer_type                AGT_TIMER
+#define _timer_index               0
+#define _timer_get_underflow_bit() R_AGT0->AGTCR_b.TUNDF
+#define _timer_clock_divider       TIMER_SOURCE_DIV_8 // dividers 1, 2 and 8 work because _timer_period is < 16-bit. the divider 4 seems not supported: acts as 1
+#define _timer_clock_freq          24000000UL
```
facchinm, line 27: This might not work on all FSP supported boards (like Portenta C33). It's very important to fix this before merging since otherwise the number of ifdefs in the core would soon become unbearable

facchinm (changes requested):

cores/arduino/time.cpp
aentinger: Yes, the defines have to go @Kees-van-der-Oord , as already mentioned [here](https://github.com/arduino/ArduinoCore-renesas/pull/52#discussion_r1265004564).

I don't see how there can be any runtime speed-up by having them in the code, as its simple string search and replace done by the preprocessor.

aentinger (changes requested):

cores/arduino/time.cpp
```diff
@@ -17,20 +19,25 @@ __attribute__((weak)) void yield() {
 }
 
 static FspTimer main_timer;
-
-static uint32_t _top_counter;
+// specifying these details as constants makes micros() faster !
```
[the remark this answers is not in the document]
Kees-van-der-Oord: dear @aentiger,

Specifying the timer constants as compiler pre-defined macros is the most CPU efficient implementation.
Using variables will add extra CPU cycles because the value of the variable must be fetched from the memory.

I timed the speed of micros() for some alternative implementations:

The test consists of calling micros() twice and comparing the value.
This is done 10000 times.

The timing results are:
                                                                average    max
Uno R3                                                    3.6 us       24 us
Uno R4 1.0.2 faulty implementation       10.0 us       17 us (omitting the negative periods)
Uno R4 macros                                        3.7 us         7 us
Uno R4 variables integer calculations       4.0 us         8 us
Uno R4 variables floating point correct  12.6 us       20 us 

Considering that the R4 clock frequency is 3x faster than the R3, it is disappointing that the call to micros() in the current implementation is 3x slower. (ignoring the overflow bug). Just fixing the overflow bug  makes it even 20% more slower.

The macros implementation just matches the R3 performance. I think this is partly because the function call overhead of the R4 is more complex than on the R3: when defining the micros() function as an inline function, the performance is much better.

The difference between 3.7 us and 4.0 us seems small, but my application makes heavily use of the micros() function, so this 10% performance improvement of the macro implementation makes a significant difference for me.

 @facchinm raised a question on the support of the Portenta C33. I compiled and tested the test sketch on the Portenta and it turns out that a different used time.cpp is at different location: ~\hardware\renesas_portenta\1.0.2\cores\arduino. I will try to find an implementation that can be used for both boards.

cores/arduino/time.cpp
Kees-van-der-Oord: Please see the comment [here](https://github.com/arduino/ArduinoCore-renesas/pull/52#discussion_r1284799160). I will try to find the most efficient implementation that also works for the Portenta C33.

Kees-van-der-Oord:
I just pushed a new version that is compatible with both the Uno R4 and the Portenta C33  (and any FSP chip with a reasonable AGT clock frequency. Optimizing the code a little more, I was able to almost match the execution time of micros() on the R4 to that on the R3:
```
              average   max
Uno R3:        3.46      24
Uno R4:        3.61       7
Portenta C33:  1.26       3

```
the code to test the micros() function:
```
void test_micros3() {
  Serial.println("testing micros3 - send input to stop");
  Serial.flush();
  uint32_t count = 0;
  uint32_t start = micros();
  uint32_t total = 0;
  uint32_t max = 0;
  uint32_t min = 0xFFFFFFFF;
  while(!Serial.available() && (count < 1000000)) {
    uint32_t t1 = micros();
    uint32_t t2 = micros();
    uint32_t period = t2 - t1;
    total += period;
    if(period > max) max = period;
    if(period < min) min = period;
    ++count;
  }
  double ave = -1;
  if(count) {
    ave = double(total) / count;
  }
  p("testing micros3 average: ");p(ave);p(" min: ");p(min);p(" max: ");p(max);Serial.println(" us");
  while(Serial.available()) Serial.read();
}
```

Kees-van-der-Oord:
I found one of reasons why the performance is so slow: the function FspTimer::get_counter(). I just pushed a version in whch the call to this function is replaced with a direct counter registry read, 
The performance is now much better: 

```
               average     max
Uno R3:        3.46 us     24 us
Uno R4:        2.12 us      5 us
Portenta C33:  0.60 us      2 us
```

cores/arduino/time.cpp
```diff
@@ -17,20 +19,28 @@ __attribute__((weak)) void yield() {
 }
 
 static FspTimer main_timer;
+const uint8_t _timer_type = AGT_TIMER;
```
aentinger, line 22: Hi @Kees-van-der-Oord ☕ 👋 

Thank you for your good work, but there has been a misunderstanding. I did not ask you to replace the macros with constants or inline functions. Instead I meant that you should directly inline them into the calling code, that is to say here:

```C++
main_timer.begin(
  TIMER_MODE_PERIODIC,
  /* _timer_type */ AGT_TIMER,
  /* _timer_index */ 0,
  _timer_period,
  1,
  /* _timer_clock_divider */ TIMER_SOURCE_DIV_8,
  timer_micros_callback);
```

cores/arduino/time.cpp
```diff
@@ -17,20 +19,28 @@ __attribute__((weak)) void yield() {
...
+static void timer_micros_callback(timer_callback_args_t __attribute((unused))* p_args) {
+	agt_time_ms += 1;
 }
 
 void startAgt() {
-	main_timer.begin(TIMER_MODE_PERIODIC, AGT_TIMER, 0, 2000.0f, 0.5f, timer_micros_callback);
-	IRQManager::getInstance().addPeripheral(IRQ_AGT,(void*)main_timer.get_cfg());
+	const uint32_t _timer_clock_freq = R_FSP_SystemClockHzGet(_timer_type == AGT_TIMER ? FSP_PRIV_CLOCK_PCLKB : FSP_PRIV_CLOCK_PCLKD);
+	_timer_period = _timer_clock_freq / ((1 << _timer_clock_divider) * 1000UL);
+	main_timer.begin(TIMER_MODE_PERIODIC, _timer_type, _timer_index, _timer_period, 1, _timer_clock_divider, timer_micros_callback);;
```
aentinger, line 40:
```suggestion
	main_timer.begin(TIMER_MODE_PERIODIC, _timer_type, _timer_index, _timer_period, 1, _timer_clock_divider, timer_micros_callback);
```

cores/arduino/time.cpp
```diff
@@ -43,10 +53,16 @@ unsigned long millis()
 }
 
 unsigned long micros() {
-
-	// Convert time to us
+	// Return time in us
 	NVIC_DisableIRQ(main_timer.get_cfg()->cycle_end_irq);
-	uint32_t time_us = ((main_timer.get_period_raw() - main_timer.get_counter()) * 1000 / main_timer.get_period_raw()) +  (agt_time_ms * 1000);
+	uint32_t ms = agt_time_ms;
+	uint32_t const down_counts = _timer_get_counter();
```
aentinger, line 59: No macro, no inline function, directly accessing the registers.
```suggestion
	uint32_t const down_counts = R_AGT0->AGT; /* Obtain the current count value of the timer. */
```

cores/arduino/time.cpp
```diff
@@ -43,10 +53,16 @@ unsigned long millis()
...
+	if (_timer_get_underflow_bit() && (down_counts > (_timer_period / 2)))
```
aentinger, line 60: Again, no macros, no inlining ... directly access the registers and maybe add a sensible comment.
```suggestion
        bool const is_underflow = R_AGT0->AGTCR_b.TUNDF != 0;
	if (is_underflow  && (down_counts > (_timer_period / 2)))
```

aentinger (changes requested):
Thank you for your work @Kees-van-der-Oord.

We are getting there, but there has been a misunderstanding between you and me. I did not ask you to replace macros with variables and inline functions, I asked you to get rid of the macros alltogether and just inline those statements directly into the code.

aentinger:
Successfully tested for
* `Portenta C33` :heavy_check_mark: 
* `Uno R4 WiFi` :heavy_check_mark: 
* `Uno R4 Minima` :heavy_check_mark:

Issue #50, opened 2023-07-14
[ArduinoEigen] Pin Definitions Conflict with Sketch/Library Variables when used with ArduinoCore-renesas


mjs513:
I posted this as well on the forum but may be better suited here: https://forum.arduino.cc/t/pin-definitions-conflict-with-sketch-library-variables/1147341

Many libraries (such as Eigen) and sketches may use variables such as A0-A5 and D0-D15 as variables versus pins definitions in pins_arduino.h:
```

#undef A0
#undef A1
#undef A2
#undef A3
#undef A4
#undef A5
static const uint8_t A0 = PIN_A0;
static const uint8_t A1 = PIN_A1;
static const uint8_t A2 = PIN_A2;
static const uint8_t A3 = PIN_A3;
static const uint8_t A4 = PIN_A4;
static const uint8_t A5 = PIN_A5;

// Digital pins
// -----------
#undef D0
#undef D1
#undef D2
#undef D3
#undef D4
#undef D5
#undef D6
#undef D7
#undef D8
#undef D9
#undef D10
#undef D11
#undef D12
#undef D13
#undef D14
#undef D15
#define D0  (0u)
#define D1  (1u)
#define D2  (2u)
#define D3  (3u)
#define D4  (4u)
#define D5  (5u)
#define D6  (6u)
#define D7  (7u)
#define D8  (8u)
#define D9  (9u)
#define D10 (10u)
#define D11 (11u)
#define D12 (12u)
#define D13 (13u)
#define D14 (14u)
#define D15 (15u)
```
So if you run a simple sketch:

```
void setup() {
  Serial.begin(115200);
  while(!Serial && millis() < 5000) {}
  int D0 = 1;
  Serial.println(D0);
  //digitalWrite(D0, HIGH);
}

void loop() {}
```

it will throw an error:
```
C:\Users\Merli\AppData\Local\Temp\.arduinoIDE-unsaved2023612-22120-liswj5.0ensn\sketch_jul12a\sketch_jul12a.ino: In function 'void setup()':
C:\Users\Merli\AppData\Local\Temp\.arduinoIDE-unsaved2023612-22120-liswj5.0ensn\sketch_jul12a\sketch_jul12a.ino:4:12: error: lvalue required as left operand of assignment
   int D0 = 1;
            ^
exit status 1
Compilation error: lvalue required as left operand of assignment
```

A more complex example would be using Eigen:
In file included from

[...]

PR #54, opened 2023-07-18
Define digital pins via `static const` instead of `#define`

mjs513:
@aentinger 
Sorry for the delay in getting this to you.  Got side tracked with a weird issue as I was trying to get the Encoder library updated so people can use with the R4 boards.  But in accordance with your request for me to submit a PR for [ArduinoEigen] Pin Definitions Conflict with Sketch/Library Variables when used with ArduinoCore-renesas Issue https://github.com/arduino/ArduinoCore-renesas/issues/50 this is the update.   As I mentioned "constexpr" throws a error so I left it as const like the analog pins.  Out of curiosity what version of c++ are the R4 using? 

regards
Mike

### DIFF
--- a/variants/MINIMA/pins_arduino.h
+++ b/variants/MINIMA/pins_arduino.h
@@ -35,6 +35,25 @@ static const uint8_t A3 = PIN_A3;
 static const uint8_t A4 = PIN_A4;
 static const uint8_t A5 = PIN_A5;
 
+// Digital pins
+// -----------
+#define PIN_D0   (0u)
+#define PIN_D1   (1u)
+#define PIN_D2   (2u)
+#define PIN_D3   (3u)
+#define PIN_D4   (4u)
+#define PIN_D5   (5u)
+#define PIN_D6   (6u)
+#define PIN_D7   (7u)
+#define PIN_D8   (8u)
+#define PIN_D9   (9u)
+#define PIN_D10   (10u)
+#define PIN_D11   (11u)
+#define PIN_D12   (12u)
+#define PIN_D13   (13u)
+#define PIN_D14   (14u)
+#define PIN_D15   (15u)
+
 #undef D0
 #undef D1
 #undef D2
--- a/variants/MINIMA/pins_arduino.h
+++ b/variants/MINIMA/pins_arduino.h
@@ -70,22 +70,22 @@ static const uint8_t A5 = PIN_A5;
 #undef D13
 #undef D14
 #undef D15
-static constexpr uint8_t D0 = PIN_D0;
-static constexpr uint8_t D1 = PIN_D1;
-static constexpr uint8_t D2 = PIN_D2;
-static constexpr uint8_t D3 = PIN_D3;
-static constexpr uint8_t D4 = PIN_D4;
-static constexpr uint8_t D5 = PIN_D5;
-static constexpr uint8_t D6 = PIN_D6;
-static constexpr uint8_t D7 = PIN_D7;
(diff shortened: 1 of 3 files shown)


aentinger:
> Out of curiosity what version of c++ are the R4 using?

C++17, see [here](https://github.com/arduino/ArduinoCore-renesas/blob/main/platform.txt#L25).

Consequently `constexpr` should not be a problem.

Mind updating the PR accordingly?, i.e. `constexpr uint8_t D0 = PIN_D0;`.

mjs513:
@aentinger - done.  Faster turn time than yesterday.  But if I try to compile this example on the minima:

```
void setup() {
  Serial.begin(115200);
  while(!Serial && millis() < 5000) {}
  int D0 = 1;
  Serial.println(D0);
  //digitalWrite(D0, HIGH);
}

void loop() {}
```
I will get this error using constexpr:

```

C:\Users\Merli\AppData\Local\Arduino15\packages\arduino\hardware\renesas_uno\1.0.2\variants\MINIMA/pins_arduino.h:75:8: error: unknown type name 'constexpr'
 static constexpr uint8_t D0 = PIN_D0;
        ^~~~~~~~~
C:\Users\Merli\AppData\Local\Arduino15\packages\arduino\hardware\renesas_uno\1.0.2\variants\MINIMA/pins_arduino.h:75:26: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'D0'
 static constexpr uint8_t D0 = PIN_D0;
```

aentinger:
Well, I get a lot of errors, but none like you ... :thinking: 

```C++
/home/alex/Arduino/hardware/arduino-git/renesas/variants/UNOWIFIR4/pins_arduino.h:54:31: error: 'PIN_D0' was not declared in this scope
 static constexpr uint8_t D0 = PIN_D0;
                               ^~~~~~
/home/alex/Arduino/hardware/arduino-git/renesas/variants/UNOWIFIR4/pins_arduino.h:54:31: note: suggested alternative: 'PIN_A0'
 static constexpr uint8_t D0 = PIN_D0;
                               ^~~~~~
                               PIN_A0
/home/alex/Arduino/hardware/arduino-git/renesas/variants/UNOWIFIR4/pins_arduino.h:55:31: error: 'PIN_D1' was not declared in this scope
 static constexpr uint8_t D1 = PIN_D1;
                               ^~~~~~
```

mjs513:
oops - darn copy and paste error.  Just pushed the fix to clear up the errors you are seeing.

mjs513:
Not home right now but will do later today.  Still saw the error in build
fails?

mjs513:
Ok - updated.

aentinger:
It seems the problem is that `Arduino.h` and by extension `pins_arduino.h` produce an error when included into a C file, i.e. [`cortex_handlers.c`](https://github.com/arduino/ArduinoCore-renesas/blob/main/cores/arduino/cortex_handlers.c) in this case.

Try renaming `cortex_handlers.c` to `.cpp` and prefix `Stacktrace_Handler `with `extern "C"`:
```diff
-void Stacktrace_Handler(void) {
+extern "C" void Stacktrace_Handler(void) {
```

aentinger:
I've reconsidered, drop `constexpr` and let's go with `const` as you've proposed since it will provide broader compatibility.

mjs513:
@aentinger
Revert back to `const` .  And Welcome back.

facchinm:
Merged, thank you!

PR #55, opened 2023-07-18
Update sineWave example of AnalogWave core library

jacobhylen:
SineWave example of the AnalogWave library was using A0 as an input pin, but the A0 pin is the DAC output pin.
https://forum.arduino.cc/t/invalid-r4-sinewave-ino-example/1149139

This PR changes the potentiometer input pin to A5, to match the other DAC examples in the library.


aentinger (approved):
LGTM 👍

Issue #44, opened 2023-07-10
Serial[x].print/write - Could use some major performance enhancement.


KurtE:
It appears like any calls to Serial.print(), including Serial.println a dn likewise Serial.write(buffer, cnt), will remain in these calls, until the last byte is queued to the underlying UART.  

Here is a simple sketch:
```
void setup() {
  Serial1.begin(115200); 
  pinMode(2, OUTPUT);
  pinMode(3, OUTPUT);
  pinMode(4, OUTPUT);
  pinMode(5, OUTPUT);
}

uint32_t loop_count = 0;
void loop() {
  digitalWrite(5, HIGH);
  Serial1.print("Hello World...");
  digitalWrite(5, LOW);
  digitalWrite(5, HIGH);
  Serial1.println(loop_count++, DEC);
  digitalWrite(5, LOW);
  delay(10);
  digitalWrite(5, HIGH);
  Serial1.write("Buffer\n", 7);
  digitalWrite(5, LOW);
  delay(1000);  
}
```
Note, I instrumented the UART::write methods in Serial.cpp:
```
/* -------------------------------------------------------------------------- */
size_t UART::write(uint8_t c) {
/* -------------------------------------------------------------------------- */  
  if(init_ok) {
    tx_done = false;
    digitalWrite(2, HIGH);
    R_SCI_UART_Write(&uart_ctrl, &c, 1);
    digitalWrite(3, HIGH);
    while (!tx_done) {}
    digitalWrite(3, LOW);
    digitalWrite(2, LOW);  
    return 1;
  }
  else {
    return 0;
  }
}

size_t  UART::write(uint8_t* c, size_t len) {
  if(init_ok) {
    tx_done = false;
    digitalWrite(2, HIGH);
    R_SCI_UART_Write(&uart_ctrl, c, len);
    digitalWrite(3, HIGH);
    while (!tx_done) {}
    digitalWrite(3, LOW);
    digitalWrite(2, LOW);
    return len;
  }
  else {
    return 0;
  }
}
```

Likewise the callback method:
```
/* -------------------------------------------------------------------------- */
void UART::WrapperCallback(uart_callback_args_t *p_args) {
/* -------------------------------------------------------------------------- */  

  uint32_t channel = p_args->channel;
  
  UART *uart_ptr = UART::g_uarts[channel];

  if(uart_ptr == nullptr) {
    return;
  }
  

  digitalWrite(4, HIGH);

  switch (p_args->event){
[...]

PR #59, opened 2023-07-18
WIP - SerialX(__UART__) tx speedup and fixes

KurtE:
Should address: #44

@facchinm, and others, would be good to get your feedback on this.  
Short version:
The current released code, outputs one character at at time and waits for the first callback, which could be because of TDRE or transfer complete, before returning.  It makes no use of the txBuffer which is defined to use 512 bytes.   If you call Serial.print/write from ISR today it hangs.  

First attempts where to continue using R_SCI_UART_Write, but there were/are difficulties in getting this to work, as it simply wants to output a buffer of size N.  Once that is done, it calls the callback to the UART code.  If I then try to start up a new transfer, in the callback, you will often lose one character.  As they will notify you of the TDRE like condition as they have already output the last character in your buffer, and if you call the R_SCI_UART_Write, it will blindly write one character out into the transfer register, but the TDRE condition is not true, so you overwrite the last character of previous buffer....

So this version does not call their write, but instead handles the TDRE like code itself...

Longer version below:
 
This is a WIP.

The released code does not use the write(buf, cnt) code, but instead unrolls it to use the one byte at a time, as the method signature did not match that in print (missing const)

Fixing this uncovered lots of issues where data would be lost as, you would typically call to output again while the last one was still putting and could lose one or two bytes.

This code bypasses the R_SCI_UART_Write(&uart_ctrl, c, len); functions and instead I do my own code wheich utilizes the txBuffer that was already part of the class.  The code both fills the buffer and if the hardware buffer has room, it takes characters from the txBuffer to fill it.

I also implemented my own version of the TDRE interrupt, to again simply take data from our txBuffer.

But to do this without changing the underlying layers, I implemented a version of attachInterruptVector, that plugs in the location of my ISR handler into the interrupt vector.

Probably could/should overwrite some of the others as well.

the SerialX.flush() code was completely busted.  But you hardly noticed it before as you were only outputting one character at a time and then waiting for an interrupt before returning.

I have some debug code still in here, hopefully disabled by a #define.

### DIFF
--- a/cores/arduino/Serial.cpp
+++ b/cores/arduino/Serial.cpp
@@ -29,7 +29,19 @@
 #undef Serial
 #endif
 
+//#define ENABLE_DEBUG
+#ifdef ENABLE_DEBUG
+#define DBGDigitalWrite digitalWrite
+#else
+inline void DBGDigitalWrite(int pin, int state) {};
+#endif
+
 UART * UART::g_uarts[MAX_UARTS] = {nullptr};
+
+// Major hacks
+extern uint32_t attachInterruptVector(uint32_t interrupt_index, void (*pfunc)());
+
+
 
 void uart_callback(uart_callback_args_t __attribute((unused)) *p_args)
 {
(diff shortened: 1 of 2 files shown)


KurtE:
Going to close this one out, and replace with different version.

No feedback, I am taking as probably not an approach that would be accepted.  So reimplemented to not change or bypass the lower layer.

PR #85, opened 2023-08-04
Update tinyusb submodule to track mainline

facchinm:
Since the patches for USB HS (Portenta C33) have been merged in mainline, we can safely abandon our fork.
Some modifications are required for changed APIs, plus HUB host support has been added for C33


maidnl (approved):
This compile out of the box and works (although I tested it only on MSD qspi example).
This new mainline version however does not solve the problem of the MSD qspi strange strings that appears when reading a file from C33.

Issue #76, opened 2023-07-31
Memory corruption in simple sketch


tttapa:
Reported at https://forum.arduino.cc/t/why-does-the-compiler-add-4x-00-to-a-string-array-element/1152890

UNO R4 Wifi, version 1.0.2 of this core.

```cpp
String patternNames[] = {
    "@@@@@",
    "Light-weight spaceship",
    "R-Pentomino",
    "Diehard",
    ""
  };

void setup() {
  Serial.begin(9600);
  delay(1000);

  Serial.println("Looking at the first characters of the first element i the String-Arrazy");
  Serial.println(String(patternNames[0].charAt(0), HEX));
  Serial.println(String(patternNames[0].charAt(1), HEX));
  Serial.println(String(patternNames[0].charAt(2), HEX));
  Serial.println(String(patternNames[0].charAt(3), HEX));
  Serial.println(String(patternNames[0].charAt(4), HEX));

  Serial.end();
}
void loop() {}
```
Output:
```
Looking at the first five characters of the first element in the String-Array
0
0
0
0
40
```
Should be 5× `40`.  
Removing the last empty string from the array causes the right characters to be printed.

I've done some reverse engineering of the binary already (https://forum.arduino.cc/t/why-does-the-compiler-add-4x-00-to-a-string-array-element/1152890/44), but could not find anything, so I suspect memory corruption by some Core function (probably a 32-bit write to the beginning of the heap, presumably somewhere in a global constructor, `initVariant`, `analogReference` or `startAgt`).  
Ideally, someone with access to an UNO R4 and a hardware debugger would debug it with the necessary watchpoints to see what causes the memory to be overwritten.

aentinger:
I suspect this is rather a problem on the usage of the various ArduinoCore-API functions ([String](https://github.com/arduino/ArduinoCore-API/blob/master/api/String.h), etc.), than a heap limitation. How about if you remove one char from a previous string? How about removing multiple?

You can just wip up some test code and execute the same test on your PC, see [here](https://github.com/arduino/ArduinoCore-API#unit-testing). All the fancy debuggers you need run on your PC 😉 .

tttapa:
Hi, thanks for the quick response.

As discussed in the forum thread, the user's code is perfectly fine.  
The problem exists elsewhere.

For good measure, I ran the following test against the ArduinoCore-API on my PC, and it passes as expected (with the sanitizers enabled and under Valgrind):
```cpp
[...]

PR #86, opened 2023-08-04
analogReference: fix heap corruption

facchinm:
on startup, `analogReference()` is being called to read the real value of 5V voltage so there's a way to retrieve the real voltage via `analogRead()`.

However, calling `R_ADC_Close` when the ADC is not yet open triggers a chain of events that ends up at

```
__STATIC_INLINE void R_FSP_IsrContextSet (IRQn_Type const irq, void * p_context) {
    gp_renesas_isr_context[irq] = p_context;
}
```

If the irq is not initialized, setting `gp_renesas_isr_context[somethig]` will overwrite random memory.

Fixes https://github.com/arduino/ArduinoCore-renesas/issues/76

### DIFF
--- a/cores/arduino/analog.cpp
+++ b/cores/arduino/analog.cpp
@@ -545,11 +545,15 @@ void analogReference(uint8_t mode) {
         ctrl = ADC_VREF_CONTROL_AVCC0_AVSS0; break;
   }
 
-  R_ADC_Close(&adc.ctrl);
+  if(adc.ctrl.opened) {
+    R_ADC_Close(&adc.ctrl);
+  }
   adc.cfg_extend.adc_vref_control = ctrl;
   R_ADC_Open(&adc.ctrl, &adc.cfg);
 
-  R_ADC_Close(&adc1.ctrl);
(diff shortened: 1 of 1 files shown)


OlafEichstaedt:
Hello @facchinm, thank you very much for finding a fix for this unpredictable behaviour. If you don't mind me asking: can you tell me how your fix will make its way into the distributed version? Will there be an update of the Uno R4 boards some day? Or is this something, that a future update of the IDE will provide.

aentinger:
>If you don't mind me asking: can you tell me how your fix will make its way into the distributed version?

The next time a new version of this core is released.

>Will there be an update of the Uno R4 boards some day? Or is this something, that a future update of the IDE will provide.

Actually, neither. Once the new core has been released it can be upgraded using [IDE](https://support.arduino.cc/hc/en-us/articles/4404691106066-How-to-update-the-core-of-your-Arduino-Board) or [arduino-cli](https://arduino.github.io/arduino-cli/0.33/commands/arduino-cli_core_upgrade/) (Technically a ArduinoCore is not part of the IDE).

Issue #88, opened 2023-08-06
RTC.begin() crashes whole processor


maxgerhardt:
This function is declared to have return value `bool`

https://github.com/arduino/ArduinoCore-renesas/blob/0e9f4df4515b508e99326e9262aaa8b17796a7ff/libraries/RTC/src/RTC.cpp#L606-L613

but there's not a single `return` keyword in that function. It ought to do `return is_initialized;` at the end.

This incorrect code causes some compilers (arm-none-eabi-gcc 10.3.1) to emit assembly that causes an infinite loop. Any sketch using `RTC.begin();` deadlocks forever. Specifically, the assembly is

![69eaa3dda954d333495272a5404b777435d6a9a7](https://github.com/arduino/ArduinoCore-renesas/assets/26485477/92a63ead-c0f7-473a-b45b-bb4c985518ba)

(Notice how there is no `pop {..., pc}` to facilitate any returning in that function. After writing `true`, the code immediately writes `false` and jumps back at the beginning of the function, repeating the cycle forever). When I fix the code

![eaed625b41282401253b94e15c48a04dd1e1fdcd](https://github.com/arduino/ArduinoCore-renesas/assets/26485477/fa1592da-656e-4f62-8c90-d8a42bf3f77f)

It does nicely return.

This code is undefined behavior which *does happpen to work* on the compiler version used in the Arduino IDE (7.0.1), however, the higher version one used in PlatformIO (10.3.1) breaks with it. (We upgraded it because, [what could possible go wrong?](https://github.com/platformio/platform-renesas-ra/issues/2#issuecomment-1625411040) Well, this).

PR #89, opened 2023-08-06
Fix undefined behavior in RTC.begin()

maxgerhardt:
Resolves compilers generating the wrong assembly per https://github.com/arduino/ArduinoCore-renesas/issues/88


aentinger (approved):
LGTM 👍 Good catch, thank you @maxgerhardt .

Interesting that it causes assembly to be produced with an infinite loop 🤔

PR #99, opened 2023-08-21
WiFi firmware 0.2.1

pennam:


pennam:
Force pushed to fix firmware submodule github url

PR #122, opened 2023-09-05
platform.txt - core platform path for tinyusb

JAndrassy:
I think the `tinyusb path` should use core platform path `{build.core.path}`, not board platform path `{runtime.platform.path}`.
the board platform can still override it in its platform.txt


JAndrassy:
the check fails are unrelated

PR #131, opened 2023-09-10
Add method to allow the GPT Timer period not to be buffered.

KurtE:
During code review of changes I made to the Servo library: https://github.com/arduino-libraries/Servo/pull/116

@iabdalkader requested that I move all of the platform specific timer changes into the core instead of being in the hardware specific sub-directory of the Servo library.

This change adds a method to tell the GPT timer, that when I make a change to the pseriod (set_period(p) ) that I want the change to happen now and not be buffered.

And updated the set_period method to check for this and directly set the not buffered register with the new period.

### DIFF
--- a/cores/arduino/FspTimer.cpp
+++ b/cores/arduino/FspTimer.cpp
@@ -459,7 +459,7 @@ bool FspTimer::set_period(uint32_t p) {
         }
     }
         else {
-            // Not buffered set it directl
+            // Not buffered set it directly
             gpt_timer->ctrl.p_reg->GTPR = p;
         }
     }
--- a/cores/arduino/FspTimer.cpp
+++ b/cores/arduino/FspTimer.cpp
@@ -11,7 +11,7 @@ bool FspTimer::force_pwm_reserved = false;
 TimerAvail_t FspTimer::gpt_used_channel[GPT_HOWMANY] = { TIMER_FREE };
 TimerAvail_t FspTimer::agt_used_channel[AGT_HOWMANY] = { TIMER_FREE };
 
-FspTimer::FspTimer(): init_ok(false), agt_timer(nullptr), gpt_timer(nullptr), type(GPT_TIMER), _buffer_period(true) {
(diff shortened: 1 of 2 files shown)


cores/arduino/FspTimer.cpp
```diff
@@ -453,9 +453,15 @@ bool FspTimer::set_period(uint32_t p) {
 /* -------------------------------------------------------------------------- */    
 
     if(type == GPT_TIMER && gpt_timer != nullptr) {
+        if (_buffer_period) {
         if (R_GPT_PeriodSet(&(gpt_timer->ctrl), p) != FSP_SUCCESS) {
             return false;
         }
+    }
+        else {
+            // Not buffered set it directl
```
iabdalkader, line 462:
```suggestion
            // Not buffered set it direct
```

cores/arduino/FspTimer.h
```diff
@@ -111,6 +112,8 @@ class FspTimer {
     bool reset();
     bool set_duty_cycle(uint32_t const duty_cycle_counts, TimerPWMChannel_t pwm_ch);
     bool set_period(uint32_t p);
+    bool use_period_buffer(bool buffer_period);
+
```
iabdalkader, line 116:
```suggestion
```

cores/arduino/FspTimer.cpp
```diff
@@ -470,6 +476,37 @@ bool FspTimer::set_period(uint32_t p) {
 
 
 
+/* -------------------------------------------------------------------------- */
+bool FspTimer::use_period_buffer(bool buffer_period) {
```
iabdalkader, line 480: I think this should be called  `set_period_buffer` just to match other functions.

```suggestion
bool FspTimer::use_period_buffer(bool buffer_period) {
```

cores/arduino/FspTimer.cpp
iabdalkader: @KurtE Thank you for the updates, could you please fix this typo as well ?

iabdalkader (approved):

PR #134, opened 2023-09-15
Fix WiFiStation buffer management

pennam:


Fashion-Corp (commented):
What was the problem? I also see that a variable, if_num, seem to be undefined, yet it is being called by reference in CnetIf.cpp, line 1376: 

uint8_t if_num;

The outcome is not in use, however, that could change :-) Well. Let me know what is the problem with the WiFIStation buffer management, and what the bugfix does.

pennam:
Sketch to reproduce the issue:

[DownloadC33-PR.zip](https://github.com/arduino/ArduinoCore-renesas/files/12648859/DownloadC33-PR.zip)

pennam:
> What was the problem? I also see that a variable, if_num, seem to be undefined, yet it is being called by reference in CnetIf.cpp, line 1376:
> 
> uint8_t if_num;
> 
> The outcome is not in use, however, that could change :-) Well. Let me know what is the problem with the WiFIStation buffer management, and what the bugfix does.

@Fashion-Corp i've updated the PR description with an explanation of the problem. `if_num` is initialized [here](https://github.com/arduino/ArduinoCore-renesas/blob/b82e6428078fb988618dfb252c4e32fc2fdad011/libraries/ESPhost/src/CEspControl.cpp#L84) if a valid mesage is found, however it is not used in `CWifiStation::task()` so it should not be an issue.

PR #147, opened 2023-09-24
WiFiS3 fix Modem timeout at ERROR response

JAndrassy:


JAndrassy:
@per1234 Hi. this on is 'imperfection'

PR #151, opened 2023-09-26
[WIRE-98] Ethernet driver and LWIP wrapper fixes

andreagilardoni:
This PR aims to fix the issues that made ethernet drivers unrealiable and unable to process complex data transfers, like file downloads. This should improve stability on all drivers using lwip wrapper.

Further improvements can be made in order to have better performances and stability, but they require a deeper knowledge of renesas fsp mechanism for RX zero copy


andreagilardoni:
This fix breaks Wifi on C33.

This fix works well with ethernet, that can leverage lwip buffer, but breaks wifi, because it uses its own queue inside of the driver. 

The wifi driver should be changed and made similar to the ethernet driver and mimic its behaviour.

andreagilardoni:
After some testing i discovered the following:
- Wifi still works as before, and these changes are compatible with #134 
- Ethernet works better than before

Drivers still need improvements, and further fine tuning of lwip library are required in order to work properly with bigger files. 

Performance needs improvement, since at the application layer we use only the <10% of the bandwidth used at the driver level. This is probably  related to:
- application layer needs to be improved
- lwip needs a better tuning in order to fit the device constraints 
- lwip wrapper has some issues that are not addressed by this PR

The following is  an example of output from an http File download of arbitrary size.
```
Application layer: 2146882/85088572 speed: 171.19 KB/s
RX bytes:    2553168, calls:       4452, failed:         98, failed alloc:         98, failed inputs:          0
TX bytes:     233394, calls:       4323, failed:          2
RX time:      62.47us speed:    2059.31KB/s
TX time:       4.26us speed:     187.03KB/s
```

Using the following ping command we can test the board with a sketch that initializes Ethernet.
```
sudo ping <BOARD_IP> -i 0.00001 -s 1450
```

Here we can see that the Driver is able to handle that packets and reach a throughput of 8592.56KB/s (67Mbit/s), and the the driver with lwip is able to handle the packets.
```
RX bytes:   67907361, calls:      45672, failed:          0, failed alloc:          0, failed inputs:          0
TX bytes:   67865334, calls:      45502, failed:         11
RX time:     124.04us speed:    8592.56KB/s
TX time:      22.60us speed:    8589.44KB/s
```

pennam (commented):
:+1:  Tested with Cloud Sketches both WiFi and Ethernet works.
:+1: Fixed file download with ethernet (no TLS)

PR #165, opened 2023-10-18
UNO WiFi R4 OTA examples: fix download return value print

pennam:


pennam:
Portenta C33 CI should not run on this library...

Issue #160, opened 2023-10-10
P214 and P215 problem


simooon442:
Hello.
On a custom minima pcb, I'm trying to read P214 and P215 with this code:

```cpp
#include <Arduino.h>
#define INPUT_PIN BSP_IO_PORT_02_PIN_15
void setup() {
  R_IOPORT_PinCfg(NULL, INPUT_PIN, IOPORT_CFG_PORT_DIRECTION_INPUT);
  Serial.begin(115200);
}

void loop() {
  bsp_io_level_t ret;
  R_IOPORT_PinRead(NULL, INPUT_PIN, &ret);
  if(ret == BSP_IO_LEVEL_LOW) {
    Serial.println("Pin LOW");
  } else {
    Serial.println("Pin HIGH");
  }
  delay(500);
}
```

But without success, ports state is only **LOW**, unable to read high level on these ports.

But when I try to read P204 for example, then I'm able to read different level state.

My question is, is the minima bootloader take into account these two port P214 and 215?

guurii:
I am also facing the same problem. Did you find any solution?

simooon442:
> I am also facing the same problem. Did you find any solution?

Hello, we solve the problem by setting parameter called "BSP_CLOCK_CFG_SUBCLOCK_POPULATED" to 0 in the bsp_cfg.h

maxgerhardt:
What magic is it that turning "subclock populated" to 0 (I guess that means "no") makes specific GPIO pins work, when different GPIO pins worked fine before this? Do you have a reference to the relevant sections in the chip's manual?

facchinm:
Pins P214 and P215 are supposed to be connected to a 32k quartz in case BSP_CLOCK_CFG_SUBCLOCK_POPULATED is set to 1; the mux is probably going to prevent the pins reconfiguration afterwards (the reference manual doesn't report anything on this topic but it looks like a sensible choice)

paulvha:
it is explained and validated at https://forum.arduino.cc/t/custom-pcb-p214-port-reading/1176644

PR #169, opened 2023-10-30
unor4: fix incorrect BSP_CLOCK_CFG_SUBCLOCK_POPULATED defines

facchinm:
Fixes https://github.com/arduino/ArduinoCore-renesas/issues/160

Special thanks to @simooon442 for spotting the issue

### DIFF
--- a/variants/MINIMA/variant.cpp
+++ b/variants/MINIMA/variant.cpp
@@ -79,6 +79,11 @@ void usb_post_initialization() {
   ((R_USB_FS0_Type*)R_USB_FS0_BASE)->USBMC_b.VDCEN = 1;
 }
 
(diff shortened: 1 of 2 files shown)


maxgerhardt:
And that requires no FSP library recompile? What code in the Arduino core listens to this macro?

simooon442:
Not able to find that. But it seems its linked to **libfsp.a** file

facchinm:
@maxgerhardt you are right, I missed some steps... I provided a new patch that doesn't involve changing libfsp, tested to work fine on UNO R4 WiFi (should work on Minima too of course).

The idea is that the function should only be called by 3rd party cores that need the pins (if they don't want to recompile libfsp)

PR #170, opened 2023-11-02
CAN Filter and Mask

Woodwarth:
Added CAN Filter and Mask

### DIFF
--- a/libraries/Arduino_CAN/examples/CANReadFilter/CANReadFilter.ino
+++ b/libraries/Arduino_CAN/examples/CANReadFilter/CANReadFilter.ino
@@ -23,8 +23,12 @@
 
 #include <Arduino_CAN.h>
 
-#define Mask11Bit 0x1FFC0000 
-#define Mask29Bit 0x1FFFFFFF 
+/**************************************************************************************
+ * CONST
+ **************************************************************************************/
+
+static uint32_t const CAN_FILTER_MASK_STANDARD = 0x1FFC0000;
+static uint32_t const CAN_FILTER_MASK_EXTENDED = 0x1FFFFFFF;
 
 /**************************************************************************************
  * SETUP/LOOP
@@ -35,8 +39,8 @@ void setup()
   Serial.begin(115200);
   while (!Serial) { }
 
-  CAN.setFilterMask_Extended(Mask29Bit);
-  CAN.setFilterMask_Standard(Mask11Bit);
+  CAN.setFilterMask_Extended(CAN_FILTER_MASK_EXTENDED);
+  CAN.setFilterMask_Standard(CAN_FILTER_MASK_STANDARD);
 
   for (int c=16; c <= 23; c++) {
     CAN.setMailboxID(c, 0x0100);
--- a/libraries/Arduino_CAN/examples/CANReadFilter/CANReadFilter.ino
+++ b/libraries/Arduino_CAN/examples/CANReadFilter/CANReadFilter.ino
@@ -1,0 +1,51 @@
+/*
+  CANRead
+
+  Receive and read CAN Bus messages
+  
+  
+  Mailbox Groups 0-3 (Mailbox  0-15) -> TRANSMIT
+  Mailbox Groups 4-5 (Mailbox 16-23) -> RECEIVE EXTENDED
+  Mailbox Groups 6-7 (Mailbox 24-31) -> RECEIVE STANDARD
+
+  See the full documentation here:
+  https://docs.arduino.cc/tutorials/uno-r4-wifi/can
+*/
+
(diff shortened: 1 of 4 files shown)


libraries/Arduino_CAN/src/R7FA4M1_CAN.cpp
```diff
@@ -187,6 +187,18 @@ void R7FA4M1_CAN::end()
   R_CAN_Close(&_can_ctrl);
 }
 
+
+void R7FA4M1_CAN::setMailboxMask(int mailbox_group, uint32_t mask)
```
aentinger, line 191:
```suggestion
void R7FA4M1_CAN::setMailboxMask(size_t const mailbox_group, uint32_t const mask)
```

libraries/Arduino_CAN/src/R7FA4M1_CAN.cpp
```diff
@@ -187,6 +187,18 @@ void R7FA4M1_CAN::end()
...
+{
+	_can_mailbox_mask[mailbox_group] = mask;
+}
+
+void R7FA4M1_CAN::setMailboxID(int mailbox, int id)
```
aentinger, line 196:
```suggestion
void R7FA4M1_CAN::setMailboxID(size_t const mailbox, uint32_t const id)
```

libraries/Arduino_CAN/src/R7FA4M1_CAN.h
```diff
@@ -54,6 +54,8 @@ class R7FA4M1_CAN final : public HardwareCAN
   bool begin(CanBitRate const can_bitrate) override;
   void end() override;
 
+  void setMailboxMask(int mailbox_group,uint32_t mask);
```
aentinger, line 57:
```suggestion
  void setMailboxMask(size_t const mailbox_group,uint32_t const mask);
```

libraries/Arduino_CAN/src/R7FA4M1_CAN.h
```diff
@@ -54,6 +54,8 @@ class R7FA4M1_CAN final : public HardwareCAN
...
+  void setMailboxID(int mailbox,int id);
```
aentinger, line 58:
```suggestion
  void setMailboxID(size_t const mailbox, uint32_t const id);
```

libraries/Arduino_CAN/src/R7FA4M1_CAN.cpp
```diff
@@ -187,6 +187,18 @@ void R7FA4M1_CAN::end()
...
+{
+	_can_mailbox[mailbox].mailbox_id = id;
```
aentinger, line 198: Use the same level of indentation for the rest of the code, that is "2 spaces" per indentation level.

libraries/Arduino_CAN/examples/CANReadFilter/CANReadFilter.ino
```diff
@@ -0,0 +1,51 @@
...
+
+void setup()
+{
+  Serial.begin(115200);
+  while (!Serial) { }
+
+  CAN.setMailboxMask(4, 0x1FFFFFFF);
+  CAN.setMailboxMask(6, 0x1FFFFFFF);
+
+  for (int c=16; c <= 31; c++) {
```
aentinger, line 33: How about using named constants instead of harcoded value?

libraries/Arduino_CAN/examples/CANReadFilter/CANReadFilter.ino
```diff
@@ -0,0 +1,51 @@
+/*
+  CANRead
+
+  Receive and read CAN Bus messages
+  
+
```
aentinger, line 6: It would be interesting to document the expectation, i.e. which messages are expected to pass through the filter, and which ones are supposed to be filtered out?

aentinger (changes requested):
Hi @Woodwarth ☕ 👋 

It's a good start! 👍 I've added a couple of suggestions to improve this PR ;)

aentinger:
Hi @Woodwarth :coffee: :wave: 

I've made the CAN filter API more Arduino friendly. Can you please test it and tell me if this still works as you want it to work?

Woodwarth:
Dont add fixed functions for setting standard/extended id's, keep it dynamic.
Added function to set id mode for mailbox (std/ext). -> will do an example for this

contant with std/ext mask should be in libary and not in example

aentinger:
>Dont add fixed functions for setting standard/extended id's, keep it dynamic.

Make a proposal. However, no internal knowledge of the library should be required. I'll be dropping your two latest commits.

aentinger (approved):

PR #175, opened 2023-11-06
WiFiS3/C3 - add legacy wl_enc_type constants aliases

JAndrassy:
previous WiFi libraries used the encryption type names ENC_TYPE_TKIP and ENC_TYPE_CCMP, which are based on the name of the encryption algorithm TKIP for WPA and CCMP for WPA2.

In WiFiS3 and WiFiC3 library constants with the well known marketing names WPA and WPA2 are used for encryption type constants.

this PR adds the legacy constants ENC_TYPE_TKIP and ENC_TYPE_CCMP  as aliases. 

EDIT: for WiFiC3 too


aentinger (approved):
LGTM :+1: Makes sense.

PR #177, opened 2023-11-06
WiFiC3 - STA BSSID getter fix

JAndrassy:
it returned own MAC. it should return BSSID of the AP it is joined to


JAndrassy:
@facchinm this is a very simple bugfix

aentinger (approved):
Static code analysis checks out okay, so :+1:

PR #178, opened 2023-11-07
RTC - Replace Serial.print()s from ISR functions in example Sketches

jacobhylen:
Currently some of the RTC examples include Serial.print() calls in callback functions that are triggered by an ISR. This is bad practice because:
- Serial.print() can depend on interrupts, which are disabled while another ISR function is running. 
- ISR functions should be as short and quick as possible.

This PR removes Serial.print() calls from callback functions in the RTC example sketches.

They are replaced by flags and conditional statements in the loops, keeping the callback functions as short as possible.


JAndrassy:
please add more context into the title. it is the issue tracker for the whole plaform

PR #184, opened 2023-11-10
WiFiC3 - macAddress() return normal bytes ordering

JAndrassy:
The firmware of the first WiFi library returned the MAC address in reversed ordering. It was fixed in examples and not in the `macAddress` method. Since then every WiFi library by Arduino reverses the MAC address in `macAddress` to reversed ordering and some reverse BSSID too. Then the examples print it reversed.

Why continue with this in new libraries? 

here BSSID was not reversed so examples printed it reversed.

overview of WiFi/Ethernet getters and setters:
https://github.com/JAndrassy/Arduino-Networking-API/blob/main/ArduinoNetAPILibs.md#network-interface-getters-and-setters

### DIFF
--- a/libraries/WiFi/examples/ConnectWithWPA/ConnectWithWPA.ino
+++ b/libraries/WiFi/examples/ConnectWithWPA/ConnectWithWPA.ino
@@ -97,14 +97,14 @@ void printCurrentNet() {
 }
 
 void printMacAddress(byte mac[]) {
-  for (int i = 5; i >= 0; i--) {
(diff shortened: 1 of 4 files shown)


aentinger:
>Why continue with this in new libraries?

Because they way it has been done before has been wrong? I'd rather fix all old libraries than "wrong" the correct implementation. What do you think?

JAndrassy:
reversed ordering is wrong. this PR does normal ordering for this library

PR #185, opened 2023-11-10
WiFiC3 example WiFiPagerServer

JAndrassy:
to demonstrate how server.available and print-to-all-clients work

other libraries including the Ethernet library have this example I created because in the other examples this is not so clear.

btw: I play with the WiFiC3, lwIWrapper and ESPHost library on an Uno R4 with an ESP32 wired on SPI


aentinger (approved):
Tested, LGTM :+1:

PR #190, opened 2023-11-15
analogReference: add getter for AR_INTERNAL if defined at variant level

facchinm:


cores/arduino/analog.cpp
```diff
@@ -570,13 +570,17 @@ float analogReference() {
       case ADC_VREF_CONTROL_VREFH0_AVSS0:
         // the user must know the voltage he applies from outside
         return NAN;
+#ifdef AR_INTERNAL_VOLTAGE
```
sebromero, line 573: The change on line 583 now requires `AR_INTERNAL_VOLTAGE` to be defined, otherwise it won't compile. Therefore this #ifdef statement is redundant.

sebromero (changes requested):
Works, but the ifdef is redundant, see comment.

sebromero (approved):
LGTM

PR #182, opened 2023-11-10
I2C hot switch master <-> slave

maidnl:
Solve: https://github.com/arduino/ArduinoCore-renesas/issues/180
This PR introduces the possibility to change back and forth the role on the I2C from Master to Slave and vice versa.

Issue #180, opened 2023-11-08
UNO R4 access to DS1307 only works if board is used as a master exclusively


Pascal2739:
Hello, for several years, I have been working with UNO R3 cards linked together by an I2C bus. Each of these cards are used alternately as a “Master” component and a “Slave” component. In my sketch, the Wire library is declared "slave" by the `Wire.begin(address)` instruction and everything works without a hitch. Recently, I purchased a UNO R4 card and it too will be integrated into my I2C network. On the I2C bus, a DS1307 real-time clock is installed which works perfectly with UNO R3 or MEGA cards. However, on the R4, access to the DS1307 only works if the card is declared as a master component exclusively (`Wire.begin()`). I tried using the `Wire.end()` function, but it had no effect. Does anyone have an idea to resolve this problem? THANKS

aentinger:
Hi @maidnl :coffee: :wave: 

Can you please take a look at this?

maidnl:
@Pascal2739 @aentinger 
The PR https://github.com/arduino/ArduinoCore-renesas/pull/182 should solve this problem (it was not possible to switch role on I2C bus). Please let give a try and let me know. Thanks.

Pascal2739:
Thank you, however, having no skills in using GITHUB (this is my first time using it), what should I do with the pieces of code? Do they replace lines of code in the library?. I'm really sorry I don't understand, but usually I never modify libraries. Thank you again for your attention.

maidnl:
@Pascal2739 no problem at all. Just wait then... the fix will be evaluated by my colleagues and if it works as expected it will be integrated (this however will take some times, so be patient). At the end you will see in the IDE you are using that an update is available, at this point if you install the update your R4 will (hopefully) work as the R3.

Pascal2739:
Thank you, I will wait, my software is under development and can wait. Thank you for everything.

PR #191, opened 2023-11-15
Allow Wire to switch from master to slave and vice versa

maidnl:
This PR solve https://github.com/arduino/ArduinoCore-renesas/issues/180 and supersedes https://github.com/arduino/ArduinoCore-renesas/pull/182.
Basically allow I2C to switch from master to slave and vice versa in a cleaner way (with less changes).

### DIFF
--- a/libraries/Wire/Wire.cpp
+++ b/libraries/Wire/Wire.cpp
@@ -383,12 +383,6 @@ void TwoWire::_begin(void) {
   irq_req.scfg = &s_i2c_cfg;
   
   if(is_master) {    
-      /* see note in the cfg_pins
-           the IRQ manager need to know the HW channel that in case of SCI 
-           peripheral is not the one in the cfg structure but the one in 
-           the Wire channel, so copy it in the request */
-      irq_req.hw_channel = channel;
-      
       if(is_sci) {
         init_ok &= IRQManager::getInstance().addPeripheral(IRQ_SCI_I2C_MASTER,&irq_req);
       }
--- a/libraries/Wire/Wire.cpp
+++ b/libraries/Wire/Wire.cpp
@@ -274,6 +274,16 @@ done:
 /* -------------------------------------------------------------------------- */
 void TwoWire::begin(void) {
 /* -------------------------------------------------------------------------- */  
+  end();
+  is_master = true;
+  _begin();
+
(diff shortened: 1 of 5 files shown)


aentinger (approved):
LGTM :+1: 

So clean ... :heart_eyes:

libraries/Wire/Wire.cpp
```diff
@@ -404,26 +410,23 @@ void TwoWire::begin(void) {
...
   /* Address is set inside begin() using slave_address member variable */
-  begin();
+  _begin();
   
 }
 
 /* -------------------------------------------------------------------------- */
 void TwoWire::begin(int address) {
 /* -------------------------------------------------------------------------- */  
-  is_master = false;
```
facchinm, line 417: Is this change necessary? What happens in case the caller requests slave -> master -> slave ?

libraries/Wire/Wire.cpp
```diff
@@ -273,6 +273,16 @@ bool TwoWire::cfg_pins(int max_index) {
 
 /* -------------------------------------------------------------------------- */
 void TwoWire::begin(void) {
+/* -------------------------------------------------------------------------- */  
+  end();
```
facchinm, line 277: Is end() needed here? Should we also add it to `begin(address)` ?

facchinm (changes requested):

libraries/Wire/Wire.cpp
maidnl: This is a simplification (I removed duplicated code). Before there were 3 different `begin(address)` functions and all of them were setting is_master = false, but in the end the all finished up to call the version `begin(uint16_t address)`. So I removed 2 lines of code from the other 2 two (master = false and slave_address = address). You can change role as many time as you want.

libraries/Wire/Wire.cpp
maidnl: end() is called in both cases -> in case of master initialization in the line you spotted, in case of slave initialization is called only in the version of begin(uint16_t address) (line 413), because the other 2 begin functions (i.e. begin(int address) and begin(uint8_t address) are in the end calling the version with uint16_t as argument.

facchinm (approved):

PR #193, opened 2023-11-17
ESPHost lib with Arduino pin functions and SPI library

JAndrassy:
ESPHost library without Renasas SDK.


JAndrassy:
I extracted the ESPHost library into a standalone repository and changed it to use the SPI library and Arduino language commands for pins and delay
https://github.com/JAndrassy/ESPHost

Issue #195, opened 2023-11-27
TwoWire::read_from and TwoWire::write_to millis rollover bug


delta-G:
The code in bot TwoWire::read_from and TwoWire::write_to includes these lines:

   ```
    timeout_ms = millis() + timeout_ms;
    while(millis() < timeout_ms && bus_status == WIRE_STATUS_UNSET && err == FSP_SUCCESS) {

    }
```
which is not safe for millis rollover.  In the rare occurrence that read_from or write_to is called within timeout_ms of rollover, the code will immediately timeout.  The AVR code for this is written using subtraction as is preached daily on the forum.  

```
  startMicros = micros();
  while(TWI_MRX == twi_state){
    if((twi_timeout_us > 0ul) && ((micros() - startMicros) > twi_timeout_us)) {
      twi_handleTimeout(twi_do_reset_on_timeout);
      return 0;
    }
  }
```

I propose fixing the code to use subtraction like the AVR code.

**Comment by JM-FRANCE** [comment]
`twi_timeout_us` is a `uint32_t`so it's always positive - you only want to check if it's not null

```
  startMicros = micros();
  while(TWI_MRX == twi_state){
    if((twi_timeout_us != 0ul) && ((micros() - startMicros) > twi_timeout_us)) {
      twi_handleTimeout(twi_do_reset_on_timeout);
      return 0;
    }
  }
```
I guess this is for efficiency reasons, avoid the call to micros() and the subtraction and compare if it's null...

maybe the compiler is smart enough to only check for 0 rather than the sign of an unsigned variable.

PR #196, opened 2023-11-27
fixing millis rollover bug x2

delta-G:
Fix TwoWire::read_from and TwoWire::write_to to use proper millis code that can handle rollover.  
Fixes issue https://github.com/arduino/ArduinoCore-renesas/issues/195

### DIFF
--- a/libraries/Wire/Wire.cpp
+++ b/libraries/Wire/Wire.cpp
@@ -478,7 +478,7 @@ uint8_t TwoWire::read_from(uint8_t address, uint8_t* data, uint8_t length, unsig
       }
     }
     uint32_t const start = millis();
-    while(((millis() - startMillis) < timeout_ms) && bus_status == WIRE_STATUS_UNSET && err == FSP_SUCCESS) {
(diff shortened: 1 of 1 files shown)


libraries/Wire/Wire.cpp
```diff
@@ -477,8 +477,8 @@ uint8_t TwoWire::read_from(uint8_t address, uint8_t* data, uint8_t length, unsig
         err = m_read(&m_i2c_ctrl,data,length,!sendStop);
       }
     }
-    timeout_ms = millis() + timeout_ms;
-    while(millis() < timeout_ms && bus_status == WIRE_STATUS_UNSET && err == FSP_SUCCESS) {
+    uint32_t startMillis = millis();
```
aentinger, line 480:
```suggestion
    uint32_t const start = millis();
```
Use const correctness ;)

aentinger (changes requested):
LGTM :+1: However, please use either "underscore case" `start_millis` or drop the `millis` suffix, since all other variables here seem to be non CamelCase.

delta-G:
I just changed the name to start as suggested.  It looks like everything compiles now.  Please let me know if it needs anything else.  

Thanks!

aentinger (approved):
LGTM :+1:

PR #197, opened 2023-11-30
OTAUpdate: do not use Error enum to return error codes

pennam:
In this way if an error code changes in the wifi firmware there is no need to change this library


pennam:
ci build of the stable version is failing because `OTA_ERROR_NONE` is still unknown, but should be safe to merge.

PR #198, opened 2023-12-05
Add gallery of icons and animations

alranel:
This PR adds a gallery of icons and animations that anyone can use :)


aentinger (approved):
LGTM :+1:

PR #199, opened 2023-12-09
WiFiC3 - config(ip) didn't set default value for DNS IP

JAndrassy:


JAndrassy:
@facchinm this little fix could be merged too. it sets the gateway IP as default DNS IP in the 'automatic' config function

PR #201, opened 2023-12-10
WiFiC3 - apply new static IP settings in _config() (a patch way)

JAndrassy:
the PR a quick-dirty-patch to show what is missing. 
a proper solution should be done by the maintainer of the library. maybe a `config` or `setAddress` method in `CNetIf`?

discovered and tested with [WiFiTest](https://github.com/JAndrassy/NetApiHelpers/tree/master/examples/WiFiTest)


aentinger:
>a proper solution should be done by the maintainer of the library

Care to provide a proper solution in form of a PR?

JAndrassy:
I can do an alternative PR but it will be less transparent

aentinger:
>I can do an alternative PR but it will be less transparent.

Let's take a look at it.

JAndrassy:
the alternative: https://github.com/arduino/ArduinoCore-renesas/pull/209

PR #205, opened 2023-12-13
Fix: compilation of "ESP32_Test.ino" fails due to non-existent function "begin".

aentinger:
I guess the function was needed earlier in the development process but that no longer seems to be the case. Also doing some small other changes in order to get the sketch to compile.

Compilation/upload on Portenta C33 via
```bash
arduino-cli compile -b arduino-git:renesas:portenta_c33 -v examples/ESP32_TEST -u -p /dev/ttyACM0
```


maidnl (approved):
Sounds good to me.

PR #207, opened 2023-12-13
lwipClient - fix mem pool leak 

JAndrassy:
fix mem pool leak in lwipClient with shared_ptr with custom deleter.
for lwipClient created from lwipServer the deleter doesn't free the struct. server releases the clients it manages. it would be better if the server used an array of lwipClient objects. then the shared_ptr in them would hold the pointer. maybe in a follow up PR if this is merged.

### DIFF
--- a/libraries/lwIpWrapper/src/lwipClient.cpp
+++ b/libraries/lwIpWrapper/src/lwipClient.cpp
@@ -8,7 +8,6 @@ extern "C" {
 
 /* -------------------------------------------------------------------------- */
 lwipClient::lwipClient()
-    : _tcp_client(NULL)
 {
 }
 /* -------------------------------------------------------------------------- */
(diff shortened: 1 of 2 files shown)


aentinger:
I'm not seeing it. Sure using a `std::shared_ptr` would be a pretty clean solution, but you could always just call `delete` in the Destructor. Having a `mem_alloc` (a C function) inside of `reset` (a member of a C++ template class) just feels super weird. What do you think?

JAndrassy:
Client must be copyable

aentinger:
> Client must be copyable

Overload copy-ctor. But yeah, I do see your point somewhat. Can you prepare a adjacent PR showing how you'd like to take this further?

JAndrassy:
> Can you prepare a adjacent PR showing how you'd like to take this further?

ok

andreagilardoni:
This issue depends on https://github.com/arduino/ArduinoCore-API/pull/218

JAndrassy:
> This issue depends on https://github.com/arduino/ArduinoCore-API/pull/218

no. `tcp_client` is not a class

PR #209, opened 2023-12-14
CNetIf - added config method to apply new static IP settings for WiFi and Ethernet

JAndrassy:
It was not possible to reconnect to network with a different static IP. IP settings worked only for first invocation of `begin` (applied in CLwipIf::get). 

in CWifi::_config `netif_set_address` call was missing. 
in Ethernet::begin static IP was not applied if `ni` was already initialized.
Now new CNetIf.config does the static IP change for both.

discovered and tested with [WiFiTest](https://github.com/JAndrassy/NetApiHelpers/tree/master/examples/WiFiTest) and [LegacyEthernetTest](https://github.com/JAndrassy/NetApiHelpers/blob/master/examples/LegacyEthernetTest/LegacyEthernetTest.ino)


JAndrassy:
I can test Ethernet now in Renesas Core so I added to this PR a fix for Ethernet which uses the same new config method.

PR #213, opened 2023-12-18
ESPHost library - protobuf-c.c as source and not as precompiled

JAndrassy:
there is no good reason to have protobuf-c.c precompiled. it compiles fast.
it was the only .o in libprotobufc.a


aentinger (approved):
LGTM :+1: You're right, there's barely a speed trade off, also now git submodule'd protobuf version matches the one compiled into your sketch.

PR #217, opened 2023-12-20
Fixed encoding of KeySlot ID.

lagosantol:
KeySlot was incorrectly encoded in a single byte.
As per https://github.com/NXPPlugNTrust/nano-package?tab=readme-ov-file#mbedtls-alt-files it should be stored over 4 bytes, in Big Endian order.


pennam (approved):
LGTM

pennam:
Thanks for the patch :slightly_smiling_face:

PR #219, opened 2023-12-20
Fixed hash algo to be selected based on input and not hardcoded

lagosantol:
The default SE05x configuration comes with 256 bit curves, so SHA256 is a better default choice.
Even better is to adapt to the input, in case the configuration is changed.


pennam:
Superseded by https://github.com/arduino/ArduinoCore-renesas/pull/229

Issue #218, opened 2023-12-20
Hardcoded digest algorithm value for SE05x is inconsistent with SE05x default configuration.


lagosantol:
https://github.com/arduino/ArduinoCore-renesas/blob/c0f158f1e0a536de76a06163bc14d19f649ba33a/extras/tls/mbedtls_alt/ecdsa_se05x.c#L171C45-L171C45

The actual algorithm to use should depend on the input size.
(20 bytes means SHA1, 32 bytes means SHA256, etc...)

pennam:
fixed with https://github.com/arduino/ArduinoCore-renesas/pull/229

PR #220, opened 2023-12-20
SSLClient SE050: fix KeySlot and signature algo configuration

### LINKED ISSUE #216: SE05x KeyID is not correctly encoded into the private key reference
The issue is in this line:

https://github.com/arduino/ArduinoCore-renesas/blob/c0f158f1e0a536de76a06163bc14d19f649ba33a/libraries/SSLClient/src/SSLClient.cpp#L380

Only the least significant byte of the key ID is stored.
The key ID should be stored in 4 bytes, big endian.

pennam:
Fixes #216  #218


pennam:
Closing as duplicate of #219 and #217

PR #222, opened 2023-12-22
Ethernet.begin - fix parameters ordering

JAndrassy:
Author of the STM32Ethernet library decided to change the ordering of the parameters for `begin` without `mac` address parameter. The EthernetC33 library is based on the STM32Ethernet library and the ordering of the parameters was not fixed.

The library is not yet used by many. It requires the Portenta C33 and a Portenta shield with Ethernet PHY module. So lets change the ordering of the parameters as soon as possible.

### DIFF
--- a/libraries/Ethernet/src/Ethernet.cpp
+++ b/libraries/Ethernet/src/Ethernet.cpp
@@ -28,29 +28,32 @@ int CEthernet::begin(unsigned long timeout, unsigned long responseTimeout) {
 /* -------------------------------------------------------------------------- */
 int CEthernet::begin(IPAddress local_ip) {
 /* -------------------------------------------------------------------------- */
-  IPAddress subnet(255, 255, 255, 0);
-  return begin(local_ip, subnet);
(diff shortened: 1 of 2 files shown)


aentinger:
Hi @pennam :coffee: :wave: 

Is the [EthernetC33](https://github.com/arduino/ArduinoCore-renesas/tree/main/libraries/Ethernet) library used anywhere  in the Arduino IoT Cloud stack yet? If not I'd be willing to merge this change.

JAndrassy:
the IoT Cloud lib uses the ConnectionHandler lib and that requires the correct ordering of parameters

https://github.com/arduino-libraries/Arduino_ConnectionHandler/blob/1c5fb9b625080b2f5b700a6693115dc005b17094/src/Arduino_EthernetConnectionHandler.cpp#L84

```
if (Ethernet.begin(nullptr, _ip, _dns, _gateway, _netmask, 15000, 4000) == 0) {
```

aentinger:
I've also checked, let's move ahead then :+1:

PR #224, opened 2023-12-23
WiFiS3 let WiFi.h include WiFiClient, WiFiUDP, WiFiServer

JAndrassy:


JAndrassy:
@pennam this one? 
as Alessandro Ranellucci wrote here https://groups.google.com/a/arduino.cc/g/developers/c/c-1I3WAfXDY/m/ZQC56DpwAgAJ
> Ideally, #include <WiFi.h> shall work out of the box on any platform (just like SPI or Wire) instead of having all the various WiFiNINA.h, ESP8266WiFi.h etc.

PR #239, opened 2024-01-12
Add WiFi Preferences support

pennam:
Needs a new [uno-r4-wifi-usb-bridge](https://github.com/arduino/uno-r4-wifi-usb-bridge)  release


pennam:
Need to add examples inside ci

PR #242, opened 2024-01-16
Add AES encrypt/decrypt and HMAC support for SE05X

pennam:
And example sketches


jowin202:
Is there already a date when the changes will be released?

PR #246, opened 2024-01-20
- fix for #244, byte-order wrong for LSBFIRST buffer transfers

RudolphRiedel:
See #244, when using buffer transfers with LSBFIRST the byte-order was wrong.


aentinger:
Hi @RudolphRiedel :coffee: :wave: 

I've verified your changes w/ this test sketch:
```C++
#include <SPI.h>

uint8_t TEST_DATA[16] = {0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15};

void setup()
{
  SPI.begin();
  SPI.beginTransaction(SPISettings(1*1000*1000, LSBFIRST, SPI_MODE0));
  SPI.transfer((void *)TEST_DATA, sizeof(TEST_DATA));
}

void loop()
{

}
```

**Before**
image

**After**
image

aentinger (approved):
Tested and verified.

aentinger:
Thank you for the fix @RudolphRiedel :coffee: :wave: :rocket:

PR #248, opened 2024-01-23
ESPHost - fix softAP password check

JAndrassy:
SoftAP password length was checked for open network and was not checked for secured

### DIFF
--- a/libraries/ESPhost/src/CCtrlWrapper.h
+++ b/libraries/ESPhost/src/CCtrlWrapper.h
@@ -729,8 +729,8 @@ public:
          cfg_ok = false;
       }
 
-      if((strlen((char *)&cfg.pwd) > MAX_PWD_LENGTH) ||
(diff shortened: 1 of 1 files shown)


pennam (changes requested):
@JAndrassy i think it should be like this

```
if ((cfg.encryption_mode != WIFI_AUTH_OPEN) && 
     ((strlen((char *)&cfg.pwd) > MAX_PWD_LENGTH) ||
      (strlen((char *)&cfg.pwd) < MIN_PWD_LENGTH))) {
```

JAndrassy:
yes. I just did a quick fix because I got error about password for WIFI_AUTH_OPEN and at the same time I released that just before that I used a short password and it took me a while to figure out why the PC doesn't want to connect.
I can't modify and retest the PR right now.

JAndrassy:
done

pennam (approved):
Thanks @JAndrassy

PR #253, opened 2024-02-01
Add software AT secure element support

pennam:
This library needs a new [uno-r4-wifi-usb-bridge](https://github.com/arduino/uno-r4-wifi-usb-bridge) release and depends on https://github.com/arduino/uno-r4-wifi-usb-bridge/pull/30

The library uses the WiFi module as a "Secure" Element sending and receiving data through AT serial interface. It only supports basic elliptic p256r1 curve operations. 

All data is sent and received as plaintext.

On the module side all operations are implemented using mbedtls software functions. Data is stored inside module NVS using the "se" namespace.

All data is stored as plaintext.


pennam:
Nedd to add examples inside CI

PR #257, opened 2024-02-12
UNOR4: rebuild libfsp with I2C patch

facchinm:
Fixes nasty bug when calling Wire.requestFrom(x, 3) and the first byte of the response was > 0x80


pillo79:
At least now it matches the comment... :cold_sweat:

PR #260, opened 2024-02-14
Fix: do not randomly clear the first line of the LED matrix.

### LINKED ISSUE #259: LED Matrix set() method does not work on top row
While testing the `set(x,y,r,g,b)` method, I noticed the `y` coordinate only works from `1` and upwards, skipping the entire top row.

The `x` coordinate works as intended (starting at `0`). 

Here is the sketch (just blinks a pixel):

```cpp
#include "ArduinoGraphics.h"
#include "Arduino_LED_Matrix.h"

ArduinoLEDMatrix matrix;

void setup() {
  //we start the LED matrix
  matrix.begin();
}

void loop() {
  int x = 0;
  int y = 0; //change to 1 to make it work
  matrix.beginDraw();
  matrix.set(x, y, 1,1,1);
  matrix.endDraw();
  delay(100);

  matrix.beginDraw();
  matrix.set(x, y, 0,0,0);
  matrix.endDraw();
  delay(100);
}
```

**Additional Note:** the `set()` method was originally created for an RGB matrix. I think it would make sense to replace `0,0,0` and `1,1,1` with something like `PIXEL_OFF` / `PIXEL_ON`, by adding a `#define PIXEL_ON 1,1,1`. 

Like: `matrix.set(x, y, PIXEL_ON)` instead of `matrix.set(x, y, 1, 1, 1)`.

aentinger:
This fixes #259.

Instead the root cause for this issue was within ArduinoGraphics, for a detailed explanation see here: https://github.com/arduino-libraries/ArduinoGraphics/pull/36 .

### DIFF
--- a/libraries/Arduino_LED_Matrix/src/Arduino_LED_Matrix.h
+++ b/libraries/Arduino_LED_Matrix/src/Arduino_LED_Matrix.h
@@ -265,10 +265,6 @@ public:
     // display the drawing
     void endDraw() {
       ArduinoGraphics::endDraw();
-      // clear first line (no idea why it gets filled with random bits, probably some math not working fine for super small displays)
-      for (int i = 0; i < canvasWidth; i++) {
-        _canvasBuffer[0][i] = 0;
-      }
       renderBitmap(_canvasBuffer, canvasHeight, canvasWidth);
     }
 


aentinger:
I've also released [ArduinoGraphics:v1.1.1](https://github.com/arduino-libraries/ArduinoGraphics/releases/tag/1.1.1) which contains the bug fix that caused the erroneous quick-fix rectified in this pull request.

PR #268, opened 2024-02-16
WiFiS3 add gettime issue 206

paulvha:
This requires a change in bridge as well.(pull request has been sent)  

In commands.h it MUST have :

#define _GETTIME                 "+GETTIME" 

regards,
Paul


paulvha:
Wifi.h and WiFi.cpp are NOT copied before compile. They are included in the pull requirest. What do I need to do ?

PR #270, opened 2024-02-16
WifiS3 add ping issue 251

paulvha:
This requires a change to bridge as well: https://github.com/arduino/uno-r4-wifi-usb-bridge/pull/47

As such it also requires in the symbolic link file `WiFiCommands.h` that has been added (included in https://github.com/arduino/uno-r4-wifi-usb-bridge/pull/47):

```cpp
#define _PINGIP                  "+PINGIP"
#define _PINGNAME           "+PINGNAME"
```


paulvha:
Wifi.h and WiFi.cpp are NOT copied before compile. They are included in the pull requirest. What do I need to do ?

Issue #206, opened 2023-12-13
WiFi.getTime() function on WiFiS3 library doesn't work


tigoe:
The getTime() function in the WiFiS2 library never gets the time via NTP. 

Here's a test:
````
#include <WiFiS3.h>            // use this for the Uno R4 WiFi board
#include "arduino_secrets.h"  // put SECRET_SSID and SECRET_PASS here

// the epoch time, obtained from the network:
unsigned long epoch = 0;

void setup() {
  // initialize serial:
  Serial.begin(9600);
}

void loop() {
  // if not connected, attempt to connect:
  if (WiFi.status() != WL_CONNECTED) {
    connectToNetwork();
  }
}

void connectToNetwork() {
  Serial.print("Attempting to connect to network ");
  Serial.println(SECRET_SSID);
  // try to connect to the network:
  while (WiFi.status() != WL_CONNECTED) {
    //Connect to WPA / WPA2 network:
    WiFi.begin(SECRET_SSID, SECRET_PASS);
    delay(2000);
  }
  // print out the device's network status:
  IPAddress ip = WiFi.localIP();
  Serial.print("IP Address: ");
  Serial.println(ip);
  // print the signal strength:
  Serial.print("Signal Strength: ");
  Serial.println(WiFi.RSSI());
  // get network time:
  getNetworkTime();
}

void getNetworkTime() {
  do {
    epoch = WiFi.getTime();
    delay(2000);
    Serial.println("NTP request attempt");
  } while (epoch <= 0);
}

````

When you run this on the Uno R4 WiFi, it'll compile, but you'll get an endless stream of "NTP request attempt" messages.

JAndrassy:
https://github.com/arduino/ArduinoCore-renesas/blob/795a81d4a48dc19ef5a2cf7cbaf8cc8e6fd2da25/libraries/WiFiS3/src/WiFi.cpp#L562-L564

tigoe:
I took the example from the WiFi libraries that Michael Margolis originally wrote, did some digging in the NTP RFC, and re-wrote it into a generic function. This should work with any of the WiFi libraries, and could also be submitted to Espressif as a pull request for the ESP32 library. I tested it there too:

````
// send an NTP request and wait for the result:
// Thanks to Michael Margolis for the original inspiration and RFC 5905 for the details. 
unsigned long getTime() {
  // NTP timestamp packet is 48 bytes:
  const int NTP_PACKET_SIZE = 48;
[...]

PR #271, opened 2024-02-21
Add `getTime` function to WiFiS3 library

paulvha:
This requires a change in the bridge firmware as well: https://github.com/arduino/uno-r4-wifi-usb-bridge/pull/48. In `commands.h` it MUST have:

```cpp
#define _GETTIME                 "+GETTIME" 
```

Fixes https://github.com/arduino/ArduinoCore-renesas/issues/206


pennam:
superseded by #394

PR #280, opened 2024-03-06
Add a clear() method to the LED matrix library

alranel:
This method makes it easy to turn off the matrix.

### DIFF
--- a/libraries/Arduino_LED_Matrix/examples/DisplaySingleFrame/DisplaySingleFrame.ino
+++ b/libraries/Arduino_LED_Matrix/examples/DisplaySingleFrame/DisplaySingleFrame.ino
@@ -34,6 +34,10 @@ void loop() {
   matrix.loadFrame(LEDMATRIX_HEART_BIG);
   delay(500);
 
+  // Turn off the display
+  matrix.clear();
+  delay(1000);
+
   // Print the current value of millis() to the serial monitor
   Serial.println(millis());
 }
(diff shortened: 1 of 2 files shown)


aentinger (commented):
Hi @alranel :coffee: :wave: 

How do you expect this API being used? If I see a `clear` API in this context I expect the RGB LED matrix to be wiped clean the moment I call it. However, that does not work. I.e. in the following example sketch both 1 and 2 are persistently displayed:

```C++
#include "ArduinoGraphics.h"
#include "Arduino_LED_Matrix.h"

ArduinoLEDMatrix matrix;

void setup() {
  matrix.begin();
}

void loop() {
  matrix.beginDraw();
  matrix.clear();
  const char text_1[] = "1";
  matrix.textFont(Font_4x6);
  matrix.beginText(0, 1, 0xFFFFFF);
  matrix.println(text_1);
  matrix.endText();
  matrix.endDraw();

  delay(1000);

  matrix.beginDraw();
  matrix.clear();
  const char text_2[] = "2";
  matrix.textFont(Font_4x6);
  matrix.beginText(8, 1, 0xFFFFFF);
  matrix.println(text_2);
  matrix.endText();
  matrix.endDraw();

  delay(1000);
}
```

alranel:
Hi @aentinger!
The intention is to provide a method to wipe the matrix instantly. I implemented it as a shortcut to `matrix.loadFrame()` called with a blank frame, which AFAIK should work. Maybe it should not be called inside `beginDraw()`? What do you recommend to achieve this goal?

alranel:
I added an example to the PR to show the intended usage.

To add some context: this is what I saw quite a few users doing manually all the time when they need to wipe the matrix. :)

PR #225, opened 2023-12-26
Return result from ArduinoLEDMatrix::begin() compatible with ArduinoGraphics.

PhilLogan:
Current implementation returns no value, resulting in annoying compiler warning - change returns value based on ArduinoGraphics docs (https://www.arduino.cc/reference/en/libraries/arduinographics/begin/).

PR #282, opened 2024-03-16
Fix return type of ArduinoLEDMatrix::begin()

mast-eu:
While playing around with the LED matrix on my Arduino R4 WiFi I noticed that the function `ArduinoLEDMatrix::begin()` causes a compiler warning:
image
(This is what I observe after setting `Preferences -> Compiler Warnings` to `All`)

The reason appears to be very simple: `ArduinoLEDMatrix::begin()` is declared to return an `int`, but never returns anything. I think this is actually ok, since the function has nothing useful it could return. 
Thus I send you this PR to change it's return type from `int` to `void`.

### DIFF
--- a/libraries/Arduino_LED_Matrix/src/Arduino_LED_Matrix.h
+++ b/libraries/Arduino_LED_Matrix/src/Arduino_LED_Matrix.h
@@ -169,14 +169,19 @@ public:
     void off(size_t pin) {
         turnLed(pin, false);
     }
-    int begin() {
+    bool begin() {
+        bool rv = true;
         uint8_t type;
         uint8_t ch = FspTimer::get_available_timer(type);
+        if(ch == -1) {
+            return false;
+        }
         // TODO: avoid passing "this" argument to remove autoscroll
-        _ledTimer.begin(TIMER_MODE_PERIODIC, type, ch, 10000.0, 50.0, turnOnLedISR, this);
-        _ledTimer.setup_overflow_irq();
-        _ledTimer.open();
-        _ledTimer.start();
+        rv &= _ledTimer.begin(TIMER_MODE_PERIODIC, type, ch, 10000.0, 50.0, turnOnLedISR, this);
+        rv &= _ledTimer.setup_overflow_irq();
+        rv &= _ledTimer.open();
+        rv &= _ledTimer.start();
+        return rv;
     }
     void next() {
         uint32_t frame[3];


facchinm:
Hi @mast-eu ,
I'd prefer to keep the `int` return and return `1` if the timer was correctly started, `0` otherwise to be coherent with the usual semantic of `begin()`.
Would you mind amending the PR in this way?

facchinm:
Could supersede #225

mast-eu:
Yes, I can amend the PR. But why do you want to keep `int`? Shouldn't `bool` be the better choice?
I'm looking for some examples of `begin()` in the code and came across
https://github.com/arduino/ArduinoCore-renesas/blob/aee54e96f74897e1af61b00be43436744ad72cf0/cores/arduino/FspTimer.cpp#L64 or https://github.com/arduino/ArduinoCore-renesas/blob/aee54e96f74897e1af61b00be43436744ad72cf0/cores/arduino/pwm.cpp#L41

facchinm:
`bool` is fine too, I'd just avoid the signature change but there's no strong reason :slightly_smiling_face:

mast-eu:
@facchinm Please have a look. I've changed the implementation to provide a meaningful return value. 

I preferred 'bool' over 'int', because maybe it's just me, but with 'int' I'd expect 0 to indicate that everything is ok and any non zero value to be an error code. While with 'bool' it's exactly the other way.

facchinm:
Super, looks much better now! Merging

misaz:
I think this change was incorect because of using bool type. Data type of begin() is actually important because begin() of ArduinoLEDMatrix overrides begin() of ArduinoGraphics and ArduinoGraphics define begin as `virtual int begin();` (https://github.com/arduino-libraries/ArduinoGraphics/blob/master/src/ArduinoGraphics.h#L41), so overriding it with method with bool output type naturally ends up with mismatch output type compiler error.

```
In file included from C:\Users\michal\AppData\Local\Temp\.arduinoIDE-unsaved2024411-121224-1myqqzw.l57b\TextWithArduinoGraphics\TextWithArduinoGraphics.ino:3:0:
C:\Users\michal\AppData\Local\Arduino15\packages\arduino\hardware\renesas_uno\1.1.0\libraries\Arduino_LED_Matrix\src/Arduino_LED_Matrix.h:172:10: error: conflicting return type specified for 'virtual bool ArduinoLEDMatrix::begin()'
     bool begin() {
          ^~~~~
In file included from C:\Users\michal\AppData\Local\Temp\.arduinoIDE-unsaved2024411-121224-1myqqzw.l57b\TextWithArduinoGraphics\TextWithArduinoGraphics.ino:2:0:
c:\Users\michal\Documents\Arduino\libraries\ArduinoGraphics\src/ArduinoGraphics.h:41:15: error:   overriding 'virtual int ArduinoGraphics::begin()'
   virtual int begin();
               ^~~~~
```

PR #287, opened 2024-04-05
define digitalPinHasPWM for UNOWIFIR4 & MINIMA

isidromv:
Defining digitalPinHasPWM will make that those sketches using it from older boards, like UNO R3 or Nano, can also be used with UNO R4, without extra complications for non advanced users (like me).

### DIFF
--- a/variants/MINIMA/pins_arduino.h
+++ b/variants/MINIMA/pins_arduino.h
@@ -90,8 +90,7 @@ static const uint8_t D13 = PIN_D13;
 static const uint8_t D14 = PIN_D14;
 static const uint8_t D15 = PIN_D15;
 
-#define digitalPinHasPWM(p)   ((p) == D3 || (p) == D5 || (p) == D6 || (p) == D9 || (p) == D10 || (p) == D11)
-
+#define digitalPinHasPWM(p) (IS_PIN_PWM(getPinCfgs(p, PIN_CFG_REQ_PWM)[0]))
 // LEDs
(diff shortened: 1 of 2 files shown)


maidnl (changes requested):
Actually there is no need to add such define because the same result can be achieved with something that is already defined in the Arduino core.
`IS_PIN_PWM(getPinCfgs(p, PIN_CFG_REQ_PWM)[0])` where `p` is the pin number (D0, D1... and so on).
So that a general macro (to be more readable) can be written as
#define digitalPinHasPWM(p)   (IS_PIN_PWM(getPinCfgs(p, PIN_CFG_REQ_PWM)[0]))

This approach is more generic because uses core functions already available, however has some drawbacks: it does not lie. 
The pin marked with ~ on the Arduino board are not in fact the only pin with PWM capability but the pin that have guarantee **independent** PWM .
Almost all available pin on Uno have PWM capability however some share the same timer and so are not completely independent
- D0 and D1 shares the same timer
- D2 and D3~ shares the same timer
- D4 and D5~ shares the same timer
- D6~ and D7 shares the same timer
- D8 and D9~ shares the same timer
- D10~ and D13 shares the same timer
- D11~ and D12 shares the same timer

isidromv:
My main point is about defining the macro to make old sketches for R3 compile.
I've updated the definition following the recommendation, thanks.

maidnl (approved):

PR #289, opened 2024-04-08
Correct name of library in file header

alrvid:
Changes "CAN library" to "FreeRTOS library".


aentinger (approved):
Thank you for the catch :+1:

PR #303, opened 2024-05-06
Modify UART Class to Use the txBuffer

delta-G:
This pull request makes changes in Serial.h and Serial.cpp so that the UART class will make use of the TxBuffer.  This greatly reduces the amount of time that `Serial.print` will block, especially for long strings.  Currently the class makes no use of the 512 bytes it has reserved for a txBuffer.  

I've changed the write functions so that they will load characters into the buffer and start a transmission if one is not already going.  

Because the HAL takes a pointer to the data, I added a char variable to read into so the buffer can be advanced.  I cannot attempt to send a larger portion of the buffer because the HAL expects contiguous memory and it will break in the case where the string to send rolls over the end of the ring buffer.  I will submit another PR later to deal with that if I find a good way. 

While it is possible to submit an entire string to the HAL at once, the string is submitted by pointer so it creates a possible race if there is some other code that changes the string before it gets fully printed.  By utilizing the txBuffer, this problem is also alleviated.  

I have tested this code in several configurations and it appears to work just dandy.  Please let me know what you think.

### DIFF
--- a/cores/arduino/Serial.cpp
+++ b/cores/arduino/Serial.cpp
@@ -334,13 +334,7 @@ void UART::flush() {
 }
 
 /* -------------------------------------------------------------------------- */
-//size_t UART::write_raw(uint8_t* c, size_t len) {
-///* -------------------------------------------------------------------------- */
-//  size_t i = 0;
-//  while (i < len) {
-//    uart_ctrl.p_reg->TDR = *(c+i);
-//    while (uart_ctrl.p_reg->SSR_b.TEND == 0) {}
-//    i++;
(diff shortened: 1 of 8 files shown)


delta-G:
I am getting the same errors on those examples without my changes.

delta-G:
Sorry for the mess.  There was some issue with submodules that created a funny commit in there.  The only files I actually changed are cores/Arduino/Serial.h and Serial.cpp

delta-G:
I'm going to close this and resubmit with the right files.

PR #309, opened 2024-05-07
Wificlient issues

andreagilardoni:
This PR aims to fix the issue for which a client class is not able to connect again after a connection loss. 
Furthermore this PR adds:
- codestyle changes to make the wificlient files consistent with themself
- code refactoring to remove undefined functions and redundant attributes
- missing rx_buffer clear inside of stop function


facchinm (approved):

JAndrassy:
in the title and in commit description it should be mentioned that it is for the WiFiS3 library. the platform has two WiFi libraries.

pennam (approved):

Issue #310, opened 2024-05-08
IRQManager doesn't implement some interrupt types


delta-G:
There are a number of peripherals available on the R4, some we have the FSP code for and some we don't.  However they all function and many require interrupts that aren't implemented in the `IRQManager` class.  

One important example of this is the Capacitive Touch Sensing Unit.  

While it is possible to write directly to the `IELSRn` registers and the vector table, the `IRQManager` has claimed responsibility for all of the slots in the ICU.  If you attach an interrupt directly there's no way to let `IRQManager` know to advance `last_interrupt_index` and so any future interrupts attached through `IRQManager` may overwrite the one you attached.  

There is a hack that I've published that involves attaching an unused interrupt and then changing the entries manually.  But I don't think this should be the standard way to do things.  

It would be difficult to add support for every possible interrupt that could be attached.  But it would be fairly simple to add a function that takes a code from table 13.4 in the [Hardware Manual](https://www.renesas.com/us/en/products/microcontrollers-microprocessors/ra-cortex-m-mcus/ra4m1-32-bit-microcontrollers-48mhz-arm-cortex-m4-and-lcd-controller-and-cap-touch-hmi) and a function pointer and sets the `IELSRn` register, puts your pointer in the vector table, and advances `last_interrupt_index`.

PR #315, opened 2024-05-13
IRQManager: implement callback based method for custom irq handlers

facchinm:
Could fix https://github.com/arduino/ArduinoCore-renesas/issues/310 

I'm not 100% convinced about the freedom <-> ease of use trade off this API should have, in the meantime it's leaning towards the freedom (since the expected audience is not the beginner).

@delta-G  what do you think?

Copying here the test sketch I added to the code
```
void irq_callback() {

}

bool config_my_funky_peripheral(unsigned int* last_interrupt, volatile uint32_t *irq_ptr, void* config) {
  funky_peripheral_cfg_t* cfg = (funky_peripheral_cfg_t*)config;
  *(irq_ptr + *last_interrupt) = (uint32_t)irq_callback;
  cfg->interrupt = *last_interrupt;
  *last_interrupt++;
  return true;
}

funky_peripheral_cfg_t funky_cfg;

void setup() {
  // put your setup code here, to run once:
  IRQManager::getInstance().addCustomPeripheral(config_my_funky_peripheral, &funky_cfg);
}

### DIFF
--- a/cores/arduino/IRQManager.cpp
+++ b/cores/arduino/IRQManager.cpp
@@ -939,6 +939,31 @@ end_config:
     return rv;
 }
 
+/*
+    The implementer should update the value of last_interrupt once done
+    eg:
+
+    bool config_my_funky_peripheral(unsigned int* last_interrupt, volatile uint32_t *irq_ptr, void* config) {
+        funky_peripheral_cfg_t* cfg = (funky_peripheral_cfg_t*)config;
+        *(irq_ptr + *last_interrupt) = (uint32_t)cfg->irq_callback;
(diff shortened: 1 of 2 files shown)


delta-G:
I could use that, but it's a little scary letting the general public have `last_interrupt_index` like that.  May I propose an alternative that keeps the code inside IRQManager?  All we need is a generic config struct that has an `IRQn_Type`, a priority, and an `elc_event_t`.  The user passes in the even type and the priority and IRQManager fills in the irq number.  

I've can put that code together real quick and push another PR and we can compare.

PR #320, opened 2024-05-18
bounds check for last_interrupt_index

delta-G:
Based on discussion from #316 I'm proposing this pull to add bounds checking for the last_interrupt_index to all functions in the IRQManager class that add interrupts.

### DIFF
--- a/cores/arduino/IRQManager.cpp
+++ b/cores/arduino/IRQManager.cpp
@@ -609,7 +609,7 @@ bool IRQManager::addPeripheral(Peripheral_t p, void *cfg) {
         mcfg->ipl = I2C_MASTER_PRIORITY;
         
         if (mcfg->txi_irq  == FSP_INVALID_VECTOR) {
-            if (last_interrupt_index + WIRE_MASTER_INTERRUPT_COUNT >= PROG_IRQ_NUM){
+            if (last_interrupt_index + WIRE_MASTER_INTERRUPT_COUNT > PROG_IRQ_NUM){
                 rv = false;
                 goto end_config;
             } 
@@ -658,7 +658,7 @@ bool IRQManager::addPeripheral(Peripheral_t p, void *cfg) {
         i2c_master_cfg_t *mcfg = (i2c_master_cfg_t *)p_cfg->mcfg;
         mcfg->ipl = I2C_MASTER_PRIORITY;
         if (mcfg->txi_irq == FSP_INVALID_VECTOR) {
-            if (last_interrupt_index + WIRE_SCI_MASTER_INTERRUPT_COUNT >= PROG_IRQ_NUM) {
+            if (last_interrupt_index + WIRE_SCI_MASTER_INTERRUPT_COUNT > PROG_IRQ_NUM) {
                 rv = false;
                 goto end_config;
             }
@@ -707,7 +707,7 @@ bool IRQManager::addPeripheral(Peripheral_t p, void *cfg) {
         scfg->eri_ipl = I2C_SLAVE_PRIORITY;
 
         if (scfg->txi_irq == FSP_INVALID_VECTOR) {
-            if (last_interrupt_index + WIRE_SLAVE_INTERRUPT_COUNT >= PROG_IRQ_NUM) {
+            if (last_interrupt_index + WIRE_SLAVE_INTERRUPT_COUNT > PROG_IRQ_NUM) {
                 rv = false;
                 goto end_config;
             }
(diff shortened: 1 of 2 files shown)


delta-G:
@iabdalkader   I mean to @ you in the description and forgot.  Does this look ok?

iabdalkader:
> @iabdalkader I mean to @ you in the description and forgot. Does this look ok?

Thanks @delta-G for sending this. This looks good, but I need to take a closer look on Monday.

cores/arduino/IRQManager.cpp
```diff
@@ -357,6 +369,11 @@ bool IRQManager::addPeripheral(Peripheral_t p, void *cfg) {
     irq_ptr += FIXED_IRQ_NUM;
     bool rv = true;
 
+    if(last_interrupt_index >= PROG_IRQ_NUM){
```
iabdalkader, line 372: Note that in the `addPeripheral` function, some peripherals require 3 or 4 IRQs and only USB checks the right number of IRQs. I think we'll need to test in every case, for example:
```C
    else if(p == IRQ_SCI_UART && cfg != NULL && ((last_interrupt_index + 4) < PROG_IRQ_NUM)) {
```

iabdalkader (changes requested):

cores/arduino/IRQManager.cpp
delta-G: Makes sense.  I'll work on it a little more tonight.

delta-G: All of the other peripherals follow a completely different pattern from USB.  In the USB section, calling with a cfg struct that already has an interrupt number results in the function returning true.  For all of the others, the function also enables the interrupts before returning true.  

Even if there are no more slots in the vector table, it's ok to enable interrupts that are already allocated there.  So I wanted to preserve the behavior and moved the check into the part of the code that actually takes new indexes. 

As an example here's what I did with the UART section:
```
else if(p == IRQ_SCI_UART && cfg != NULL) {
        uart_cfg_t *p_cfg = (uart_cfg_t *)cfg;
        
        if (p_cfg->txi_irq == FSP_INVALID_VECTOR) {
            if (last_interrupt_index + UART_INTERRUPT_COUNT > PROG_IRQ_NUM){
                rv = false;
                goto end_config;
            }
            /* TX interrupt */
            p_cfg->txi_ipl = UART_SCI_PRIORITY;
            p_cfg->txi_irq = (IRQn_Type)last_interrupt_index;
            *(irq_ptr + last_interrupt_index) = (uint32_t)sci_uart_txi_isr;
            set_sci_tx_link_event(last_interrupt_index, p_cfg->channel);
            last_interrupt_index++;

            /* TX-ERROR interrupt */
            p_cfg->tei_ipl = UART_SCI_PRIORITY;
            p_cfg->tei_irq = (IRQn_Type)last_interrupt_index;
            *(irq_ptr + last_interrupt_index) = (uint32_t)sci_uart_tei_isr;
            set_sci_tei_link_event(last_interrupt_index, p_cfg->channel);
            last_interrupt_index++;

            /* RX interrupt */
            p_cfg->rxi_ipl = UART_SCI_PRIORITY;
            p_cfg->rxi_irq = (IRQn_Type)last_interrupt_index;
            *(irq_ptr + last_interrupt_index) = (uint32_t)sci_uart_rxi_isr;
            set_sci_rx_link_event(last_interrupt_index, p_cfg->channel);
            last_interrupt_index++;

            /* RX-ERROR interrupt */
            p_cfg->eri_ipl = UART_SCI_PRIORITY;
            p_cfg->eri_irq = (IRQn_Type)last_interrupt_index;
            *(irq_ptr + last_interrupt_index) = (uint32_t)sci_uart_eri_isr;
            set_sci_eri_link_event(last_interrupt_index, p_cfg->channel);
            last_interrupt_index++;
        }
        R_BSP_IrqEnable (p_cfg->txi_irq);
        R_BSP_IrqEnable (p_cfg->tei_irq);
        R_BSP_IrqEnable (p_cfg->rxi_irq);
        R_BSP_IrqEnable (p_cfg->eri_irq);
    }
```

The only place we want to return false (assuming we want to maintain the current behavior in other scenarios) is if `txi_irq == FSP_INVALID_VECTOR` AND there are not enough interrupts left.  However it's not enough alone that `txi_irq` is set to something else to make the function return false before the changes, so I made sure it wouldn't be the case after the changes.  

I don't necessarily know if I'm a fan of the pattern there.  I think I would have written it more like the USB code.  But I don't know what other code relies on this behavior so I didn't want to alter it.  

Also note that I'm using `>` where it really feels like I need a `>=`.  I really thought about what someone might think looking at that and almost put the `=` there anyway.  But it's right with just `>`.  It's one of those "fence-post" "off-by-one" type things.  

Because the index currently held in `last_interrupt_index` is actually the first one to be used, it counts when we add them up.  So if `last_interrupt_index` is currently at 28 and we need to attach 4 interrupts, they will go in slots 28, 29, 30, and 31 and those are all legal.  But `28 + 4 == 32`.  So equals 32 is ok, it's only greater than 32 that's a problem.

It feels wrong, but I wrote it out on paper to make sure.  I'm certain it's right.

iabdalkader (approved):

PR #322, opened 2024-05-20
 Fix Inconsistent Sleep Current by relocating the Ethernet Clock AGT Timer 

cristidragomir97:
**Description:**
This pull request addresses an issue with inconsistent sleep current consumption on the board. Previously, the board would sometimes draw around 100µA and other times around 2mA. After thorough investigation, it was determined that the issue was due to the lowest power supply state (LPM_POWER_SUPPLY_DEEPCUT3) of the Renesas RA6M5 chip turning off the AGT (Asynchronous General Purpose Timer), which conflicted with the AGT timer used for the Ethernet clock defined in variant.cpp.

**Changes:**
- Removed the AGT timer definition for the Ethernet clock in variant.cpp.
- Wrapped the AGT timer functionality in the EthernetClock class
- Added timer instantiation in the begin method of the Ethernet class.
- Added timer deletion in the disconnect method of the Ethernet class.

**Testing:**
- Verified that the board enters a consistent low-power state with current consumption around 60µA.
- Ensured that Ethernet functionality is not impacted by the changes.

**Additional Notes:**
- This fix improves power efficiency and stability in low-power modes.
- Please review the changes to ensure they align with the overall project architecture and coding standards.

### DIFF
--- a/libraries/Ethernet/src/Ethernet.cpp
+++ b/libraries/Ethernet/src/Ethernet.cpp
@@ -1,5 +1,5 @@
 #include <EthernetC33.h>
-
+#include <EthernetClock.h>
 /*
  * The old implementation of the begin set a default mac address:
  * this does not make any sense.
@@ -12,9 +12,14 @@
 /* -------------------------------------------------------------------------- */
 int CEthernet::begin(unsigned long timeout, unsigned long responseTimeout) {
 /* -------------------------------------------------------------------------- */  
+    
+  ethernetTimer = new EthernetClock();
+  ethernetTimer->start();
+  delay(2);
   (void)responseTimeout;
 
   int rv = 0;
+
 
   ni = CLwipIf::getInstance().get(NI_ETHERNET);
   if(ni != nullptr) {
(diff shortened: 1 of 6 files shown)


libraries/Ethernet/src/EthernetClock.h
```diff
@@ -0,0 +1,25 @@
+
+#ifndef ETHERNET_CLOCK_H
+#define ETHERNET_CLOCK_H
+
+
+#include "FspTimer.h"
+
+#define AGT_TIMER_CHANNEL 3
+#define ETHERNET_CLK_PIN  BSP_IO_PORT_06_PIN_00
```
facchinm, line 9: I'd move these defines in the variant folder (in `pins_arduino.h`), since the library might also run on HW that doesn't use the same pins.

libraries/Ethernet/src/Ethernet.cpp
```diff
@@ -56,6 +62,11 @@ int CEthernet::begin(IPAddress local_ip, IPAddress dns_server, IPAddress gateway
 int CEthernet::begin(IPAddress local_ip, IPAddress dns_server, IPAddress gateway, IPAddress subnet) {
 /* -------------------------------------------------------------------------- */  
   
+  this -> ethernetTimer = new EthernetClock();
```
facchinm, line 65: I'd surround the clock initialization (and deinitialization) between `#ifdef ETHERNET_CLK_PIN` to take into account possible future platform that use a different method to provide the 25Mhz clock

facchinm (changes requested):

PR #323, opened 2024-05-21
Add compiler.xxx.extra_flags to receipes in platform.txt

isidromv:
Adding extra flags for compilation (e.g. global defines) is possible via `compiler.xxx.extra_flags` in receipes for other platfomrs (AVR, SAMD).
But this is not possible for this platfomr since this variable is not used in the receipe definition.


facchinm:
Merged, thanks!

PR #326, opened 2024-05-27
WifiS3 added method to get uint32 version

andreagilardoni:
This functionality depends on https://github.com/arduino/uno-r4-wifi-usb-bridge/pull/54


pennam (approved):
:muscle:

PR #328, opened 2024-05-28
Fix for Serial flush() returning before transmission has completed

Hsubtnarg:
Problem: flush() would return when the transmission buffer was empty, but the txBuffer was not being used so flush() would return straight away. write() would wait until it had sent all bytes to the UART before returning, but there would still be bytes in the UART that had not been sent, so it would return 1-11 bytes early

Fix: Used the UART_EVENT_TX_COMPLETE call back to determine when the UART has finished transmitting and use this to determine when flush() should return

### DIFF
--- a/cores/arduino/Serial.cpp
+++ b/cores/arduino/Serial.cpp
@@ -58,12 +58,16 @@ void UART::WrapperCallback(uart_callback_args_t *p_args) {
       {
           break;
       }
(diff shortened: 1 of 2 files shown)


pennam:
Hi @Hsubtnarg thanks for your contribution. I've squashed and move your commits to https://github.com/arduino/ArduinoCore-renesas/pull/405

PR #331, opened 2024-06-02
Update USB.cpp for removing delays that is interfering to get high polling rate

skhrlx:
Removed delay permitting higher polling rate mouses

Some users use this while gaming, and, today, gaming mouses can easily reach 8000mhz of polling rate, and setting a delay of 10, will stuck any mouse at 125hz


cuccune:
As for the ard leonado r3, can you make it go up to 1000hz?

skhrlx:
> As for the ard leonado r3, can you make it go up to 1000hz?

Not from my interest now, but, it's open-source too (you can take a look on it), and, on uc probably must have an thread about it, since valo scene is using it for years

PR #335, opened 2024-06-09
Clearing canvas buffer in ArduinoLEDMatrix::clear() in addition to frame buffer

stefanmarks:
In the 1.2.0 update, the override of ArduinoLEDMatrix::clear() only cleared the frame buffer, but not the canvas buffer (Issue #336).
This resulted in pixels staying on in animations using matrix.beginDraw(), matrix.set(x, y, r, g, b), matrix.endDraw() even after a call to clear().


stefanmarks:
Addresses issue #336

PR #348, opened 2024-07-11
Fix in formula for creating frequency from MIDI note

x-1xps:
The stated formula for converting midi-notes to frequency is wrong – the base and exponent of the second factor should be interchanged


JAndrassy:
add more context to the title of the PR. the current title doesn't make sense in the context of the whole Renesas platform repo

per1234 (commented):
The equation is correct as written:

https://en.wikipedia.org/wiki/Piano_key_frequencies

x-1xps:
@per1234 please check again.. my correction reflects the formula stated in the wikipedia article.

per1234 (approved):
Thanks for the fix @x-1xps

PR #357, opened 2024-08-06
Update Ethernet.cpp

rowingdude:
Using `const` liberally so that the compiler can try to optimize this down a bit more. 

I put in a couple nullptr checks to help stability as I've had my controller crash, hopefully this helps someone.

I created a `configure_static_ip` function so that `int CEthernet::begin ...` can remain function specific.

Set `ni` to a nullptr to avoid any potential "use after free" issues.

### DIFF
--- a/libraries/Ethernet/src/Ethernet.cpp
+++ b/libraries/Ethernet/src/Ethernet.cpp
@@ -30,23 +30,6 @@ int CEthernet::begin(unsigned long timeout, unsigned long responseTimeout) {
     return 0;
 }
 
-/* -------------------------------------------------------------------------- */  
-int CEthernet::configureStaticIP(const IPAddress& local_ip, const IPAddress& dns_server, const IPAddress& gateway, const IPAddress& subnet) {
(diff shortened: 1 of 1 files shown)


JAndrassy (commented):
could you please move `configureStaticIP' bellow the `begin` which had that code so it is compared in the diff

rowingdude:
Yes, sorry. I got side-tracked adding those annoying /*------ ... --*/s

rowingdude:
I will go hunt through the libraries/Wifi/Examples and update as needed to get things moving. I feel like I'm going to end up porting all of the newer WiFi code over to Arduino.

rowingdude:
I will do that, thank you for your feedback, these big projects are new to me and I'm learning as I go.

rowingdude:
I'm going to close this until I work out the bugs. Thank you, initially my local repo didn't have Ethernet cloned into it and I was going off the web version, however, with the new update, I cannot submit this as is. Cheers

Issue #321, opened 2024-05-20
RTC date set incorrectly for years before 2000


NathanielJS1541:
## The Bug 🐛

Using an Arduino Uno R4 WiFi, if the RTC date is set to before 2000-01-01, it will incorrectly report the date when it is read back from the RTC. For an easy repro, see the sketch I have provided in the [Repro Sketch section](#repro-sketch).

## Contents

This has gotten pretty long, so I'll leave this here as an easy way to jump between sections...

- [The Bug 🐛](#the-bug-)
- [Contents](#contents)
- [Debugging Attempt...](#debugging-attempt)
  - [My Findings](#my-findings)
  - [Issue Summary](#issue-summary)
- [Repro Sketch](#repro-sketch)

## Debugging Attempt...

During my search for answers, I have spent a long time trying to work out where it's all going wrong. I will do my best to impart that knowledge onto others here.

### My Findings

The RTC in the RA4M1 has a 100 year calendar, which in itself is not a problem, but it seems to be bodged into working with the C standard for `tm_year` (defined as the number of years since 1900 in [the UNIX specification](https://pubs.opengroup.org/onlinepubs/7908799/xsh/time.h.html)).

In the Arduino RTC library, the years of RTCTime get 1900 subtracted from them, so 1970 becomes 70 and 2017 becomes 117 etc:

```cpp
#define TM_YEAR_OFFSET   (1900)

...

bool RTCTime::setYear(int _y) {
    if (_y >= TM_YEAR_OFFSET) {
        _y -= TM_YEAR_OFFSET;
    }
    year = _y;
    stime.tm_year = _y;
    //stime.tm_yday = day + yday(year, Month2tm(month));
    return true;
}
```

I am unable to debug what happens within the fsp, but according to the comments validation of the year should fail for 70 since the expected range is 100-199. This is detailed in [r_rtc.c](https://github.com/renesas/fsp/blob/d5e6aff684490605063cf5289bad0531cffdc175/ra/fsp/src/r_rtc/r_rtc.c#L1328-L1333) of [Renesas/fsp](https://github.com/renesas/fsp). It looks like the validation is disabled, i.e. `RTC_CFG_PARAM_CHECKING_ENABLE` is not defined, since the result of `RTC.setTime` for these invalid dates returns `True`.

[...]

Issue #294, opened 2024-04-27
WL Definitions Missing


MatthewA1:
Several definitions found in [`wl_definitions.h`](https://github.com/arduino-libraries/WiFi/blob/master/src/utility/wl_definitions.h) for the normal [Arduino WiFi library](https://github.com/arduino-libraries/WiFi) are missing in the Renesas Core's WiFiS3 library's [`WiFiTypes.h`](https://github.com/arduino/ArduinoCore-renesas/blob/main/libraries/WiFiS3/src/WiFiTypes.h). I'm not sure all of the definitions are needed since this implementation is different than the normal WiFi library, but at least some of them are.
For example, many of the length definitions such as `WL_MAC_ADDR_LENGTH` are helpful when declaring arrays in sketches.

PR #358, opened 2024-08-07
Update digital.cpp

rowingdude:
Add a conditional statement to check if we're on an Uno, this flips IOPORT_CFG_PMOS to _NMOS as needed.

### DIFF
--- a/cores/arduino/digital.cpp
+++ b/cores/arduino/digital.cpp
@@ -1,29 +1,33 @@
 #include "Arduino.h"
 
 void pinMode(pin_size_t pin, const PinMode mode) {
-	switch (mode) {
-		case INPUT:
-		case INPUT_PULLDOWN: // TODO: document the INPUT_PULLDOWN is unavailable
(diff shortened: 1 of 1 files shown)


rowingdude:
Per issue #321, I have also edited the RTC.cpp file with two if-statements in `setYear` and `setUnixTime` to ensure the date range is valid per the RTC manufacturer's guidelines (year 2000 to 2099).

rowingdude:
The initial commit (digital.cpp) is to address issue #343

rowingdude:
Per issue #294 I integrated the wl_definitions.h `#defines` with the new `wifiTypes.h` header file.

rowingdude:
There seem to be some errors in the original syntax of the rtc.cpp file. I will investigate them.

rowingdude:
I have attempted to address the "qualified-id in declaration" which is done by removing the RTClock:: in `bool RTClock::isSet()` so it's just `bool isSet() { ... }`.

rowingdude:
I have rolled back the changes due to more breakage in the Portenta, Minima, and UnoR4Wifi codebases. I'll look into why those are occuring. The PR is now versioned at my **4d6cd7** commit.

PR #362, opened 2024-08-18
True 16 bit transfer, and new loop friendly 16 bit transfer

drmcnelson:
This replaces the transfer16() with a true 16 bit transfer, adds a new loop friendly 16 bit transfer, and rename the old transfer16() as two bytes to transfer16_asbytes()

### DIFF
--- a/libraries/SPI/SPI.cpp
+++ b/libraries/SPI/SPI.cpp
@@ -3,12 +3,15 @@
  * Copyright (c) 2014 by Paul Stoffregen <paul@pjrc.com> (Transaction API)
  * Copyright (c) 2014 by Matthijs Kooijman <matthijs@stdin.nl> (SPISettings AVR)
  * Copyright (c) 2014 by Andrew J. Kroll <xxxajk@gmail.com> (atomicity fixes)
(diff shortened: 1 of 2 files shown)


per1234 (commented):
Thanks for taking the time to submit a pull request @drmcnelson.

Arduino has a strict policy that we can only accept pull requests after the contributor has signed the contributor license agreement. Since you have stated in https://github.com/arduino/ArduinoCore-renesas/issues/363 that you refuse to sign the agreement, we can not accept this pull request.

drmcnelson:
In fact I did sign the CLA,  I filled in the form, ticked the box and clicked the button.

Your software after that asked for access to my private data.   When a bot asks for access to private data, that is where I draw the line, as any normal person should.

I plan to keep the fork up, or at least the patched SPI support with the functioning and improved 16 bit transfers.     It seems there are people who want it.

Or, you could incorporate into the main tree yourself, you certainly have my permission to do that.  Just let me know if you do.

drmcnelson:
By the way "invalid" is insulting.   The problem as described above, is in CLA assistant.   And if you check around you will see complaints about this are common.

PR #365, opened 2024-08-19
Exclude OPTA_* from unor4 package

pennam:
I suspect OPTA folder should be excluded from uno r4 package


facchinm:
Indeed, but should also be excluded from the uno_r4 package :thinking: The hacky solution would be to also add `.uno_r4_only` but it's not very explicit (nor nice)

pennam:
the `.unor4_only` flag is already in place. so the OPTA variant in excluded from portenta package and included in uno r4. Adding also .portenta_only would exclude the opta varian to be included for both.

PR #368, opened 2024-08-21
Add missing return statement in SyncCanMsgRingbuffer::isFull()

Wetmelon:


aentinger (approved):
Good catch! Thank you @Wetmelon :rocket: :bow: .

PR #400, opened 2024-11-28
Initialize local variable for AP config

alrvid:
The AP configuration variable wasn't initialized, so the fields contained random data. Then, they were filled using memcpy() with the size of the data to copy, leaving the remainder of the fields containing random data. For example, the pwd variable containing the entered password was copied without null termination into the pwd field, and the rest of the pwd field contained random data. This caused connection failures depending on the content of the random data.


sebromero (approved):
LGTM

Issue #406, opened 2024-11-30
Non-blocking version of  WiFi::localIP()


ondra-novak:
Can we have non-blocking version of the function localIP?

I have WDT set on 5 seconds and this function causes WDT to trigger

https://github.com/arduino/ArduinoCore-renesas/blob/03b4a91db49dae7601701c33cd09924ea6b363d3/libraries/WiFiS3/src/WiFi.cpp#L336

Thank you

JAndrassy:
after WiFi.begin, the localIP() should return immediately because WiFi.begin waits for DHCP. how do you initialize WiFi?

ondra-novak:
I am also setting timeout to zero, because wifi connection is not mandatory to my device. Its priority is to control connected devices. The control cycle has 1 millisecond resolution. I can handle aprx 20mills a common modem-esp32 exchange, but i cannot handle 5 seconds blocking call.

MYusufY:
I made PR that fixes this by adding `localIPNonBlocking()` function.

**PR:**
- #408 

_I hope it fixes your issue and have a nice day :)_

JAndrassy:
>  am also setting timeout to zero

but then don't call localIP() if there is no WiFi connoection

ondra-novak:
> > am also setting timeout to zero, because wifi connection is not mandatory on my device
> 
> but then don't call localIP() if there is no WiFi connoection

You don't understant. My controller running in the main loop where many operations are scheduler depend on various conditions such a time, sensors or status of various components. One component is Wifi connection. So the task calls WiFi.begin(), then peridodically checks for status and then , when status is OK, retrieves localIP. 

now consider environment with bad wifi connection. The WiFi status is OK, but immediatelly after the wifi connection is lost and localIP blocks. Or other scenario, when WiFi.status is OK, but probably DHCP handshake is still in progress. As I understand, this is the main reason, why the function is blocking. So non-blocking version can simply return "still in progress" and we are good, the Task object can postpone the operation until next cycle.

MYusufY:
> > I made PR that fixes this by adding `localIPNonBlocking()` function.
>> 
> > _I hope it fixes your issue and have a nice day :)_
> 
> Thank you, I already happily forked this PR. 


Glad you solved it! Have a nice day.

MYusufY:
> correct solution is change the esp32 firmware to not wait for localIP()

You mean using **ESP-IDF**?

But i think solution is solution.

[...]

PR #408, opened 2024-11-30
Add Non-blocking Version of localIP Function for WiFiS3 library and fix FifoBuffer.h

MYusufY:
This PR introduces a non-blocking version of the localIP function, named localIPNonBlocking, to the WiFi.cpp file. The new function addresses situations where the current localIP implementation, which uses a blocking delay(100) call, can cause issues such as triggering a watchdog timer (WDT).

This PR also solves:
- #406

### New Function: `localIPNonBlocking`
Added to libraries/WiFiS3/src/WiFi.cpp at line 357 (after the existing localIP implementation).
Utilizes a timeout mechanism with millis() to avoid blocking delays.
Returns the device's local IP address or a default address (0.0.0.0) if the timeout expires.

### Example Usage:
```
unsigned long timeout = 5000; // 5-second timeout
IPAddress ip = WiFi.localIPNonBlocking(timeout);

if (ip) {
    Serial.println(ip); // Print the IP if successfully retrieved
} else {
    Serial.println("Failed to get IP within the timeout period");
}

```

### DIFF
--- a/cores/arduino/FifoBuffer.h
+++ b/cores/arduino/FifoBuffer.h
@@ -27,7 +27,6 @@
 
 namespace arduino {
 
-
 #define FIFO_DEFAULT_SIZE 64
 
 template <typename T, uint32_t size = FIFO_DEFAULT_SIZE>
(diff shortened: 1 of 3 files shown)


JAndrassy:
the localIP() function just gets the IP from the esp32 over UART. it should be quick

this platform has more networking libraries so please add to the title that this is about WiFiS3 library

cores/arduino/FifoBuffer.h
andreagilardoni, line 1: Do you have any good reasons to make changes to this file? If not I would recommend you to drop them and open a new PR with these changes instead.

libraries/WiFiS3/src/WiFi.h
andreagilardoni, line 1: I think it is better to override the functionality of `localIP()` instead of writing a new one. I think you had a good Idea to specify in the function signature the timeout, so I would keep it, but with the addition of making it default to `_timeout` value. Could you also modify the comment above accordingly?

andreagilardoni:
I would also change the PR name, since we are changing the behavior from an attempt based approach to a timeout based one. The function is still blocking, but now the user may set the timeout to 0 to make it non blocking and deal with the logic of retrial on his own.

MYusufY:
Thanks for your interest! I will try to make these changes and create a new PR soon. Also, i think i accidentaly edited that file :)

andreagilardoni:
Don't worry, I can guide you through this and fix this PR, you don't need to open a new one for this feature. 

The usual method to contribute on github is to create a branch on your personal fork of the repo you are contributing to. That way you can always go back to the public repo status with `git checkout`. Please tell me if you need any help on that.

pennam:
Superseded by https://github.com/arduino/ArduinoCore-renesas/pull/450

PR #414, opened 2024-12-09
Allow FreeRTOS hook for C33 only and improve its reliability

facchinm:
Fixes concerns from https://github.com/arduino/ArduinoCore-renesas/pull/369#issuecomment-2519980477

@pennam @hexnet1234 @gpb01

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -12,7 +12,7 @@ portenta_c33.build.fpu=-mfpu=fpv5-sp-d16
 portenta_c33.build.float-abi=-mfloat-abi=hard
 
 portenta_c33.build.board=PORTENTA_C33
-portenta_c33.build.defines=-DF_CPU=200000000
+portenta_c33.build.defines=-DF_CPU=200000000 -DPROVIDE_FREERTOS_HOOK
 portenta_c33.vid.0=0x2341
 portenta_c33.pid.0=0x0068
 portenta_c33.vid.1=0x2341
--- a/cores/arduino/main.cpp
+++ b/cores/arduino/main.cpp
@@ -113,7 +113,9 @@ void arduino_main(void)
 #endif
    startAgt();
    setup();
+#ifdef PROVIDE_FREERTOS_HOOK
    start_freertos_on_header_inclusion();
+#endif
    while (1)
    {
       loop();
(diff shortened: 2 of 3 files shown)


gpb01:
@facchinm: Perhaps it would be better to condition, with the same #ifdef, also the code within port.c so as to have it compiled **only if it is really useful**. This, for those who use in IDE a "_Compiler warning_" level higher than the minimum, could avoid a series of unnecessary warnings during compilation for UNO R4. :-)

Cordially,

Guglielmo

hexnet1234:
I agree with @gpb01.

If I remember correctly, the built-in compiler of the Arduino IDE does strip unused functions out of the binary,
but adding this check seems like a good idea. Just to get rid of warnings, and (very slightly) improve compile time.
Compile-time improvements like this are small on their own, but do add up eventually.

Best regards
hexnet1234

gpb01:
> If I remember correctly, the built-in compiler of the Arduino IDE does strip unused functions out of the binary

Yes, correct!

> Just to get rid of warnings,

The compiler gives the following warning:

```
/Users/gpb01/Desktop/Arduino Portable 1.3.1.app/Contents/Java/portable/packages/arduino/hardware/renesas_uno/1.3.1/libraries/Arduino_FreeRTOS/src/portable/FSP/port.c: In function 'loop_thread_func':
/Users/gpb01/Desktop/Arduino Portable 1.3.1.app/Contents/Java/portable/packages/arduino/hardware/renesas_uno/1.3.1/libraries/Arduino_FreeRTOS/src/portable/FSP/port.c:238:9: warning: implicit declaration of function 'loop'; did you mean 'bool'? [-Wimplicit-function-declaration]
         loop();
         ^~~~
         bool
```

... and, since it is very easy to remove it, why not do it? :-)

Guglielmo

facchinm:
Done in [b83fd20](https://github.com/arduino/ArduinoCore-renesas/pull/414/commits/b83fd201289ab7925d9062064f605dcee2372d20) , if the CI turns green I'm going to merge @pennam

gpb01:
@facchinm: Thanks! :-)

Guglielmo

hexnet1234:
Looks good 👍 

Best regards
hexnet1234

Wetmelon:
I get a warning for loop() now on C33 1.3.2 and an error if I don't define it for the idle task, is that intended?

PR #428, opened 2025-02-04
[C33] Kvstore

andreagilardoni:
Importing KVStore from mbedos to be used with portenta c33

### DIFF
--- a/.github/workflows/compile-examples.yml
+++ b/.github/workflows/compile-examples.yml
@@ -76,6 +76,7 @@ jobs:
               - libraries/RTC/examples/RTC_NTPSync
               - libraries/RTC/examples/RTC_Alarm
               - libraries/SFU
+              - libraries/KVStore/examples/StartCounter
           - board:
               fqbn: "arduino-git:renesas:portenta_c33"
             additional-sketch-paths: |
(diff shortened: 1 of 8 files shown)


libraries/KVStore/KVStore.h
```diff
@@ -0,0 +1,230 @@
...
+ * limitations under the License.
+ */
+
+#ifndef MBED_KVSTORE_H
+#define MBED_KVSTORE_H
+
+#include <stdint.h>
+#include <string.h>
+
+// FIXME define mbed errors adequately
```
pennam, line 23: What is the problem with the following defines? or the comment is outdated ?

pennam (changes requested):
If this code is intended to work on C33 only you need to add the `.portenta_only` file to avoid this folder to be included in the packaging of the  UNO R4 core.


Is there a reason why there are no examples and library.properties files ?

libraries/KVStore/KVStore.h
andreagilardoni: I forgot about that. The point is that mbedos has a particular way of defining errors and make them reusable and work across all the different modules it has. I hardcoded the ones we care about for kvstore. 

Do we care about having the complete definition of error codes? or should we just use the hardcoded ones?

pennam: We can go with the hardcoded ones, but we have to check there are no clashes with the other imported mbed modules like the ones in the `BlockDevices` and `Storage` folders

pennam (approved):

PR #433, opened 2025-02-10
WiFiS3: Add timeout for localIP

### LINKED ISSUE #406: Non-blocking version of  WiFi::localIP()

pennam:
Adds timeout for localIP() instead of waiting a fixed amount of time. Other proposed solution https://github.com/arduino/ArduinoCore-renesas/pull/408

Fixes #406

### DIFF
--- a/libraries/WiFiS3/src/WiFi.cpp
+++ b/libraries/WiFiS3/src/WiFi.cpp
@@ -333,7 +333,6 @@ IPAddress CWifi::localIP() {
 
    unsigned long start_time = millis();
    do {
-      delay(100);
       if(modem.write(string(PROMPT(_MODE)),res, "%s" , CMD_READ(_MODE)))  {
          if(atoi(res.c_str()) == 1) {
             if(modem.write(string(PROMPT(_IPSTA)),res, "%s%d\r\n" , CMD_WRITE(_IPSTA), IP_ADDR)) {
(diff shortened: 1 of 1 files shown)


JAndrassy:
WiFi.begin is blocking until DHCP gets an IP address. there is no reason to have a timeout in localIP(). it will only return immediately if WiFi.begin was successful or wait the complete timeout if WiFi.begin didn't run or didn't connect.

pennam:
@JAndrassy I get your point. I do not understand why in the original code localIP() was doing the retry loop so I opted for a less impacting change.  Probably @maidnl has some hints about the original reason of the retries.

maidnl:
@pennam if I remember correctly the problem was related to the use of `begin()` and, immediately after, `localIP()`. In this case the ESP32 took some (little) time to set up the address and give it back so that the first call could fail and give the impression of a failed connection or wrong DHCP set up. For this reason I added this "retry" mechanism. I think is safe to keep some sort of retry.

maidnl (approved):

pennam:
Superseded by #450

PR #436, opened 2025-02-12
C33 SSL client fix stop() and connect( .. ) with psk

pennam:
Also suppress some othe minor warnings


maidnl (approved):
Seems good to me, but I am not expert on this topic...

PR #459, opened 2025-03-21
WiFiS3 library: allow selection of WIFI_MODE_APSTA mode

cotestatnt:
I am developing a WebServer library that includes a WiFi manager. 
In the case of DHCP connections, it is essential to provide the user with the IP address assigned by the router.

In this situation (and in other scenarios), it is necessary to have the ability to operate in both WiFi Access Point mode and Station mode, simultaneously, after successfully connecting to the SSID.

### DIFF
--- a/libraries/WiFiS3/src/WiFi.cpp
+++ b/libraries/WiFiS3/src/WiFi.cpp
@@ -42,11 +42,11 @@ int CWifi::begin(const char* ssid) {
 }
 
 /* -------------------------------------------------------------------------- */
-int CWifi::begin(const char* ssid, const char *passphrase) {
+int CWifi::begin(const char* ssid, const char *passphrase, uint8_t mode) {
 /* -------------------------------------------------------------------------- */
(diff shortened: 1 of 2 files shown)


JAndrassy:
but which IP does then WiFi.localIP() return ?

cotestatnt:
In my pull request, the `localIP()` method, in case of WIFI_MODE_APSTA, returns the address assigned by the router, but in fact it would be more correct to provide a second method that explicitly returns the IP address of the AP as happens with the ESP32 core for Arduino calling the [IPAddress softAPIP()](https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/src/WiFiAP.h#L98C3-L98C24) method, while I think `localIP()` should always return the STA address.

cotestatnt:
Regarding your doubt about the WiFi channel, I don't think it's a problem and in fact as you can see from this screenshot the WiFi manager works as expected.

image

PR #497, opened 2025-10-08
Add timeout handling for I2C in Wire

HanzHager:
Adding timeout handling for I2C in Wire according to previously defined solution on other platforms
Also reduced the default timeout from 1000ms till 25ms
Also created src directory as well as adding keywords.txt and library.properties so that the "Wire" part of the repository can be handled as a separate library


HanzHager:
This should solve the issue #492

